PR #26, opened 2024-08-02
docs: add nix to install methods

donovanglover:
Added in https://github.com/NixOS/nixpkgs/pull/330527


mierak:
Hey, thank you for doing this

The changes seem good, just one question, I do not know much about nix, but shouldn't we note somewhere that the unstable channel has to be used?

donovanglover:
Done. Will be in stable around November/December when the next release comes out.

mierak:
Great, thank you

Issue #29, opened 2024-08-05
Cant run with wezterm


cc8dea2e-4a29-4394-af6d-af0ccc73647a:
When running rmpc using wezterm, I get the message `Error: No such file or directory (os error 2)`.
Runs only if set `image_method: None` in config.ron.

Backtrace:
```
Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /home/user_name/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/backtrace.rs:27:14
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/result.rs:1989:27
   2: rmpc::utils::tmux::is_passthrough_enabled
             at ./src/utils.rs:108:22
   3: rmpc::utils::image_proto::is_kitty_supported
             at ./src/utils.rs:171:17
   4: rmpc::utils::image_proto::determine_image_support
             at ./src/utils.rs:137:12
   5: rmpc::config::ConfigFile::into_config
             at ./src/config/mod.rs:157:48
   6: rmpc::main
             at ./src/main.rs:120:25
   7: core::ops::function::FnOnce::call_once
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/ops/function.rs:250:5
   8: std::sys_common::backtrace::__rust_begin_short_backtrace
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/sys_common/backtrace.rs:155:18
   9: std::rt::lang_start::{{closure}}
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/rt.rs:159:18
  10: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/ops/function.rs:284:13
  11: std::panicking::try::do_call
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/panicking.rs:559:40
  12: std::panicking::try
             at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/panicking.rs:523:19
  13: std::panic::catch_unwind
[...]

PR #34, opened 2024-08-08
Only test tmux passthrough if TERM_PROGRAM=tmux

TheOnlyMrCat:
I tried running this on ghostty, and got the same error as #29.

I don't have tmux installed, and the error was occurring when trying to spawn a `tmux` process.

Ghostty sets `TERM_PROGRAM=ghostty`, so it detected a running tmux even though I wasn't running tmux. This fixes that.


mierak:
Looks good to me, thanks. I actually found this issue myself yesterday when implementing the iterm2 image protocol so I want to work on detecting the correct protocol anyway.

PR #35, opened 2024-08-09
feat: include git info in version and log

mierak:


src/main.rs
```diff
@@ -85,7 +85,11 @@ fn main() -> Result<()> {
             std::io::stdout().write_all(include_bytes!("../assets/example_theme.ron"))?;
         }
         Some(Command::Version) => {
-            println!("rmpc version: {}", env!("CARGO_PKG_VERSION"));
+            println!(
+                "rmpc {} ({})",
+                env!("VERGEN_GIT_DESCRIBE"),
+                env!("VERGEN_GIT_COMMIT_DATE"),
```
donovanglover, line 91: Would it be possible to remove `VERGEN_GIT_COMMIT_DATE` here? Makes packaging harder for nixpkgs since rmpc builds without git and this would require either manual date changes or the addition of an update script.

donovanglover: Currently I workaround this by changing the string to an arbitrary value ("nixpkgs" in this case) but I want to make sure you're okay with that https://github.com/NixOS/nixpkgs/pull/334026/commits/0f77bfb83a20f6228b85f31801323a6af4264baa

mierak: I do not mind removing the date itself. But wouldn't that mean getting rid of all the git related stuff? Meaning the `VERGEN_GIT_DESCRIBE` as well as the emmiter in `build.rs`? I would really prefer having commit info in the version itself so no confusion of built version can be made

donovanglover: In nixpkgs we set the `VERGEN` environment variables ourselves since packages are built without `.git` directories. It's easy to set `VERGEN_GIT_DESCRIBE` to the version of the package being built, but we don't have an easy way to get the commit date.

When these variables aren't set, it results in `rmpc version` returning `rmpc VERGEN_IDEMPOTENT_OUTPUT (VERGEN_IDEMPOTENT_OUTPUT)`. My patch changes this to `rmpc 0.3.0 (nixpkgs)`, and the `version` will always include the date for untagged commits (e.g., `0.3.0-unstable-2024-08-19`).

donovanglover: If you're okay with the version being like this then you don't have to change anything. If it's possible to make `VERGEN_GIT_COMMIT_DATE` not print the `()` part if it doesn't exist that'd be cool too, but I suppose you'd miss out on the `(nixpkgs)` identifier.

PR #40, opened 2024-08-14
fix: remove commit date to help nix pkg

mierak:


donovanglover:
LGTM thanks. Please don't feel like you have to change things just for nixpkgs!

I think it's possible to keep the date if you really wanted to. Looks like upstream suggests to use [`fail_on_error()`](https://github.com/rustyhorde/vergen/issues/174#issuecomment-1495914817) and [`option_env!`](https://github.com/rustyhorde/vergen/issues/299#issuecomment-2106151621) is also an option.

mierak:
This particular case is completely fine by me to remove. Do you want to release the patch version or is it fine like this?

Also a good learning experience on how nix handles things for me. I know about `option_env!` but did not realize that nix builds without git.

PR #75, opened 2024-09-28
docs: update Arch Linux instructions

orhun:
Now packaged in the official repositories: <https://archlinux.org/packages/extra/x86_64/rmpc/> 🥳


mierak:
Oh wow! thank you for this.

PR #121, opened 2024-11-01
fix(ui): skip album art for stream songs

soifou:
When the song is a stream (eg. a webradio), the file URI is not a local path but an HTTP URL.

Depending on the computer you run rmpc, there's a lag displaying the Queue view because it needs to query mpd about the album art. And by providing an URL it fails everytime obviously.

When I say lag, the elapsed time may vary from 500ms to over 2s sometimes on a potatoe hardware (and feel instant on a brand new machine), here is one of the log entry of my beloved potatoe:

```
2024-11-01T17:59:27.762451247Z DEBUG src/ui/panes/album_art.rs:78 message="Searching for album art" file="https://icecast.radiofrance.fr/fiphiphop-hifi.aac"
2024-11-01T17:59:29.011874642Z WARN  src/mpd/proto_client.rs:160 message="Expected binary data but got 'OK'"
2024-11-01T17:59:29.011928583Z DEBUG src/mpd/mpd_client.rs:484 message="No album art found, falling back to placeholder image"
2024-11-01T17:59:29.011952774Z DEBUG src/ui/panes/album_art.rs:80 message="Found album art" elapsed="1.249503322s" size="None"
```

This PR fallback directly to the default cover if the song file looks like a protocol instead of a path.
I refactor the logic a bit since it was called twice. Seems cleaner to me.

This is probably not the best approach but it works. I chose to check if it looks like protocol but maybe only http?s is necessary. I did not dug that much. And I'm still very new to Rust so feel free to adapt or even close this PR if you have other idea on how to handle it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1583,6 +1583,7 @@ dependencies = [
  "itertools 0.13.0",
  "log",
  "ratatui",
+ "regex",
  "ron",
  "rstest",
  "rustix",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ sysinfo = "0.32.0"
 color_quant = "1.1.0"
 enum-map = "2.7.3"
 textwrap = "0.16.1"
+regex = "1.11.1"
 
 [build-dependencies]
 clap = { workspace = true }
--- a/src/ui/panes/album_art.rs
+++ b/src/ui/panes/album_art.rs
@@ -7,6 +7,7 @@ use crate::{
 };
 use anyhow::Result;
 use ratatui::{layout::Rect, Frame};
+use regex::Regex;
 
 use super::Pane;
 
@@ -35,6 +36,23 @@ impl AlbumArtPane {
             ),
         }
     }
+}
+
+fn fetch_album_art(client: &mut impl MpdClient, context: &AppContext) -> Result<Option<Vec<u8>>> {
+    if !matches!(context.config.album_art.method.into(), ImageProtocol::None) {
+        if let Some(current_song) = context.queue.iter().find(|v| Some(v.id) == context.status.songid) {
+            // Skip if it looks like a protocol (e.g., a webradio)
+            let protocol_pattern = Regex::new(r"^[a-z]+://").expect("Failed to create protocol regex");
+            if !protocol_pattern.is_match(current_song.file.as_str()) {
+                let start = std::time::Instant::now();
+                log::debug!(file = current_song.file.as_str(); "Searching for album art");
+                let result = client.find_album_art(current_song.file.as_str())?;
+                log::debug!(elapsed:? = start.elapsed(), size = result.as_ref().map(|v|v.len()); "Found album art");
+                return Ok(result);
+            }
+        }
+    }
+    return Ok(None);
 }
 
 impl Pane for AlbumArtPane {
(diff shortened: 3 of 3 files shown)


mierak:
I am not against this, but is there really no case where album art can be read by MPD which are not local files? MPD supports various [input plugins](https://mpd.readthedocs.io/en/latest/plugins.html#input-plugins). For example SMB, is it not possible for album arts to work in there and this functionality would break this maybe? Im unsure about any of this since I dont use them.

Just throwing some ideas: Maybe this could be configurable, either something like `disable_for_remote_files: bool` or `disabled_protocols: &str[]`. The latter would avoid adding the Regex crate as well.

soifou:
Thanks for your input! Yes it needs to be rethink.

I can absolutely conceive mpd is capable of fetching album art from various protocols as nfs or smb... deny all protocols is a stupid idea. But over http, really? I need to delve deeper into the doc to find out more. For now, I put the PR in draft.

mierak:
I just tested a remote song with embedded album art. 

MPD will display album art with no issues, you can try with this song for example `rmpc add "https://puu.sh/KhSfj/4f80cc93d0.opus"`. 

The fun part, rmpc has default binary limit per one iteration of the download loop set to (I think) 4096 bytes which results in an excessive amount of network trips for a single album art download. If we set the binary limit to something more reasonable for the modern age, it is actually reasonably quick. You can try it on master vs this branch: the [binary-limit](https://github.com/mierak/rmpc/tree/feat/binary-limit) branch from this PR https://github.com/mierak/rmpc/pull/122

soifou:
Oh that's interesting indeed, I've vaguely head about opus but never used it. Metadata and covert art! Nice!

For info mpd set its binary size to [8192 bytes](https://github.com/MusicPlayerDaemon/MPD/blob/acb9ee9792dcdcecbe5f0736332678996fd7510b/src/client/Client.hxx#L77)
 
It's night and day between the 2 branches but still it takes at least 2sec to display the embedded art... 

The only solution I see right now would be to fetch it in a separate thread and retrigger the cover art.

soifou:
> The only solution I see right now would be to fetch it in a separate thread and retrigger the cover art.

As well as increase the binarylimit as you did for sure

mierak:
> Oh that's interesting indeed, I've vaguely head about opus but never used it. Metadata and covert art! Nice!

It would work for other formats too.

I agree that separate thread is the right solution. Problem is that the way the MPD client is currently set up does not allow for it easily. This could be worked around by creating a new separate client for the download thread. I do not like that but am not against it per se.

Also I really need to do another refactor pass on the album art code. it is quite a mess.

soifou:
> It would work for other formats too.

This does not work for streams coming from icecast/shoutcast servers:
- icecast: `rmpc add "https://icecast.radiofrance.fr/fip-hifi.aac"`
- shoutcast : `rmpc add "http://stream.zuccaradio.com:80/stream"`

AFAIU those servers can share metadata but on demand, the client must include `Icy-Metadata: 1` header in the GET request. [Source](https://web.archive.org/web/20191121035806/http://www.smackfu.com/stuff/programming/shoutcast.html). Also found this [rust lib](https://github.com/aschey/icy-metadata). But I don't even know how this could be handled in rmpc. 

Anyway you convince me that this PR was on a wrong road, closed! Thanks again for your time!

mierak:
> This does not work for streams coming from icecast/shoutcast servers:

I mean other audio formats like mp3 not streams

I will try to solve this issue, the ui thread being blocked while in download is quite annoying.

mierak:
Just a small update, this is a bit worse than I thought. There is obviously quite a few inefficiencies on rmpc's side, but the UI still locks up even after moving the download to a separate thread.

The reason for this is MPD itself, it slows down to a crawl when its fetching remote album arts. 

Simple way test is:
```bash
rmpc clear # careful, this will clear the current queue
rmpc add "https://puu.sh/KhSfj/4f80cc93d0.opus"
rmpc play
rmpc albumart --output /dev/null
```
And then run ncmpcpp and try to do anything at all, or even directly execute protocol commands via something like ncat.

We could move the whole interfacing with MPD to a separate thread, but thats a very big change and would bring another set of difficulties.
Disabling the album arts for HTTP might be the correct solution after all...

soifou:
Thanks for the update, I honestly didn't put much thought of it since the last time, but if MPD is the problem, might be worth to open an issue to find out what they think about it? For now I agree we can skip the fetch until we see a viable solution. The project is still young and I doubt this impact a lot of users for now. 

In the meantime I'm trying to get more familiar with your codebase because I had some ideas to enhance the `add` functionality. I took some time yesterday to experiment and I can now add one or several songs to the queue, external or within the library. It accepts recursive directories. The CLI accept path completion and wildcards. I rely on the MPD `decoders` command to see if the song can be added (and thus decoded), discarding the others. And of course, I implemented the CLI+modal `decoders`. All of this with no extra libs. Still need some work and tests but I think the cli+modal stuff is ready. If you're intereted let me know, will create some separate (draft) PRs.

mierak:
I would like to look into the issue a bit more before opening issue at MPD. But I will disable album arts for http/https for now.

If you have any questions about some mess of a code you find in there dont hesitate to ask me. Anyway, sure you can create a PR.
Also just a note, MPD cannot access local files unless rmpc is connected via local socket so that should probably be handled.

PR #127, opened 2024-11-03
feat: disable album art per protocol

mierak:


mierak:
@soifou With this you should be able to disable album art per protocol/input plugin. By default http and https are disabled, what do you think?

soifou:
Yes looks good to me! And the refactor does not harm :)

PR #128, opened 2024-11-04
feat: enhance queue deletion confirm modal

soifou:
This change improves the usability and clarity of the queue deletion confirmation modal.

- Renamed buttons from "Save" and "Cancel" to "Confirm" and "Cancel" for clarity
- Added support for TAB to cycle through 'Confirm' and 'Cancel' buttons
- Changed the Up and Down key for the Left and Right instead
- Updated the confirmation text to include a warning about the action being irreversible
- Removed unnecessary title for a cleaner UI

### DIFF
--- a/src/ui/modals/confirm_queue_clear.rs
+++ b/src/ui/modals/confirm_queue_clear.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use ratatui::{
-    prelude::{Constraint, Layout, Margin},
+    layout::Alignment,
+    prelude::{Constraint, Layout},
     style::Style,
     symbols::{self, border},
     widgets::{Block, Borders, Clear, Paragraph, Wrap},
@@ -8,7 +9,7 @@ use ratatui::{
 };
 
 use crate::{
-    config::keys::CommonAction,
+    config::keys::{CommonAction, GlobalAction},
     context::AppContext,
     mpd::{client::Client, mpd_client::MpdClient},
     shared::{
@@ -38,7 +39,7 @@ pub struct ConfirmQueueClearModal<'a> {
 impl ConfirmQueueClearModal<'_> {
     pub fn new(context: &AppContext) -> Self {
         let mut button_group_state = ButtonGroupState::default();
-        let buttons = vec![Button::default().label("Save"), Button::default().label("Cancel")];
+        let buttons = vec![Button::default().label("Clear"), Button::default().label("Cancel")];
         button_group_state.set_button_count(buttons.len());
         let button_group = ButtonGroup::default()
             .active_style(context.config.theme.current_item_style)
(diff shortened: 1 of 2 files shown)


src/ui/modals/confirm_queue_clear.rs
```diff
@@ -38,7 +40,7 @@ pub struct ConfirmQueueClearModal<'a> {
 impl ConfirmQueueClearModal<'_> {
     pub fn new(context: &AppContext) -> Self {
         let mut button_group_state = ButtonGroupState::default();
-        let buttons = vec![Button::default().label("Save"), Button::default().label("Cancel")];
+        let buttons = vec![Button::default().label("Confirm"), Button::default().label("Cancel")];
```
mierak, line 43: I think it should say "Clear" instead to name a specific action instead of generic confirm.

src/ui/modals/confirm_queue_clear.rs
```diff
@@ -120,7 +116,11 @@ impl Modal for ConfirmQueueClearModal<'_> {
                 }
                 _ => {}
             }
-        };
+        } else if key.code() == KeyCode::Tab {
```
mierak, line 119: I like the idea, but not the implementation. I would like to minimize behavior not explicitly bound to a key. Maybe Next/PreviousTab could be used here? It is a bit obscure but those two binds should never do anything when a modal is open.

src/ui/modals/confirm_queue_clear.rs
```diff
@@ -60,51 +62,45 @@ impl ConfirmQueueClearModal<'_> {
...
+        frame.render_widget(paragraph, content_area);
         frame.render_stateful_widget(&mut self.button_group, buttons_area, &mut self.button_group_state);
         Ok(())
     }
 
     fn handle_key(&mut self, key: &mut KeyEvent, client: &mut Client<'_>, context: &mut AppContext) -> Result<()> {
         if let Some(action) = key.as_common_action(context) {
             match action {
-                CommonAction::Down => {
+                CommonAction::Right => {
```
mierak, line 98: What do you think about keeping both Right | Down and Left | Up?

src/ui/modals/confirm_queue_clear.rs
soifou: Gosh, I didn't pay too much attention to the actions until now, thanks for the pointer! I'll change it.

src/ui/modals/confirm_queue_clear.rs
soifou: No objection!

src/ui/modals/confirm_queue_clear.rs
mierak: No real substantial reason, really. I initially put it to Up/Down as that effectively allowed to cycle through the whole list and both the buttons in the add to playlist modal with one button. Though having it on Right/Down only makes more sense I guess. Its up to you, Im fine with either way.

src/ui/modals/confirm_queue_clear.rs
```diff
@@ -120,7 +115,47 @@ impl Modal for ConfirmQueueClearModal<'_> {
                 }
                 _ => {}
             }
-        };
+        } else if let Some(action) = key.as_global_action(context) {
```
soifou, line 118: @mierak This is this what you meant, right?

src/ui/modals/confirm_queue_clear.rs
soifou: I you don't mind, I prefer to keep it like this for now, if someone complains, we put back in the blink of an eye

src/ui/modals/confirm_queue_clear.rs
mierak: Yup. `stop_propagation` should not be needed now since the event gets consumed automatically by `as_global_action`. You can also use the default branch to not have to list all the unused binds.

soifou: ah ok that's good to hear :+1:

mierak:
thanks!

PR #130, opened 2024-11-04
fix: weird rendering and click to select in modal tables with wrapped lines

mierak:
Not the ideal solution as this effectively makes a single table row into multiple ones, but should not matter for this use case.


soifou:
Yeah that's fine like this :+1:

PR #137, opened 2024-11-09
feat: add confirmation modal for playlist deletion

soifou:
Similar to the queue list deletion behavior, a playlist can contain a **lot** of songs; hence, the confirmation modal. 

This prevents users like me to fat-fingering this action and delete its beloved playlist ;)


mierak:
Looks good to me, thanks!

PR #150, opened 2024-11-14
refactor: confirm modals into single generic one

mierak:
With this we can have confirmation modals on demand without needless boilerplate


soifou:
Nice, I wasn't sure about the purpose of this `confirm_modal` while browsing the code to do the playlist modal.

mierak:
Yeah, It was just unfinished for a while

PR #151, opened 2024-11-15
Add a (temporary) try on NixOS 24.05 instruction in the docs.

Piecuuu:
This commit is intended to be reverted in, I think, about ~3-4 months from 24.11's release, that should be enough time for everyone to make the switch. This PR is for the people that are/will still be on 24.05.


mierak:
merged, thanks

Issue #160, opened 2024-11-21
Scrolling through playlists is slow


Holzhaus:
### Description

When navigating through the playlists in the playlists tab using the arrow keys, the entire applications starts to lag. 

### Steps to reproduce

When holding and then releasing the down arrow key, the application keeps going through the playlists for some time and is unresponsive to user input. To reproduce, use a very slow MPC server, e.g. and old RPi with a slow external HDD :sweat_smile: 

### Expected behavior

ncmpcpp is quite fast in comparison. I assume the main reason is that the actual loading of the playlist content happens in the background and the request is canceled when the user switches to another playlist before the current one was loaded.

### What terminal emulator are you running rmpc in?

alacritty 0.14.0 (22a44757)

### rmpc debuginfo

<details>
  <summary>rmpc debuginfo</summary>

  ```sh
rmpc 0.6.0 git Arch Linux

Config path          /home/jan/.config/rmpc/config.ron
Theme path           None

MPD:
Address              "127.0.0.1:6600"
Resolved Address     IpAndPort("raspberrypi-4b.lan:6600")
MPD_HOST             raspberrypi-4b.lan
MPD_PORT             unset

Youtube playback:
Cache dir            None
ffmpeg               installed       ffmpeg version n7.1 Copyright (c) 2000-2024 the FFmpeg developers
ffprobe              installed       ffprobe version n7.1 Copyright (c) 2007-2024 the FFmpeg developers
yt-dlp               not installed   Version not available
python3              installed       Python 3.12.7
python-mutagen       installed       PRESENT

Image protocol:
Requested            Auto
Resolved             Unsupported
TMUX                 false
ueberzugpp           not installed   Version not available  
```
</details>
alacritty

### Config file

default config

### Relevant log output

n/a

mierak:
Yeah that is an issue with how the update loop is setup in rmpc: `keypress -> push to queue -> read from queue -> start rendering frame -> request data from mpd -> finish render`. There is no way to cancel/discard in flight requests at the moment, but I agree that has to change somehow. 

We could sort of get around this by quickly caching the results instead of refetching on every change, but it would not solve the issue for the first pass through the playlists.

Holzhaus:
Awesome.

Here's what pressing and holding the up/down buttons on the main branch look like (yes, it really hands that long, i didn't release the button until after I stopped the recording):

https://github.com/user-attachments/assets/2c74015a-06a3-4490-a463-e993b3bd4e28

[...]

Issue #147, opened 2024-11-12
Ordering albums in artist tab


FLchs:
### Description

Hello it's me again,
In artists and album artist panes, albums are ordered alphabetically.
Ordering them by release date makes more sense to me, I know I could add the release date to the folder name but I think it clutters the view.

### Proposed solution

Afaik MPD protocol doesn't allow  to get an ordered or "rich" list from the list command so it is not as trivial as it seems.

I tried to make a proof of concept of a solution but it makes browsing the library quite slower as I have to get all the songs of an artist and then process it rather than getting a list of album names from MPD... 
I think ncmpcpp add some kind of delay before rendering the album column as a workaround.

So here is my POC, be aware I'm by no way familiar with Rust so I'm 100% sure that if you find this request pertinent you will find a more elegant way.
https://github.com/mierak/rmpc/compare/master...FLchs:rmpc:feat/order-album-by-date

FLchs:
I put some more thoughts in this issue and I may have a better idea:
Allowing (via config/theme) to format the Artist/Album/Tracks column, including some kind of trim to remove year/track number/anything else ?

For example, with a library structured like this :
```
Library
└── ArtistA
    ├── 2023 - Album1
    │   ├── 01 - Song1
    │   ├── 02 - Song2
    │   └── ...
    └── 2024 - Album2
        └── ...
```
rmpc will display a table looking like this:
|  ArtistA | 2023 - Album1  | 01 - Song1  | 
|---|---|---|
|   | 2023 - Album2  | 01 - Song2  | 
|   |   | 01 - Song3  | 
|   |   | ...  | 

But adding support for library formatting (and another property)  :
```
library_table_format :
        rows: [
            (
                artist: [
                    (style: (modifiers: "Bold", fg: "red")),
                ],
                album: [
                (style: (modifiers: "TrimDate") // Weird ?
                        triml: 6 // remove the first 6 chars ?
                        regex: "^\d{4} - " // fancy/best :) )
                ],
                track: [
                    (style: (fg: "blue"), regex: "^\d{2} - "),
                ]
            ),
        ]
```
would render:
rmpc will display a table looking like this:
|  **ArtistA** (bold,red) | Album1  | Song1  | 
|---|---|---|
|   | Album2  | Song2  | 
|   |   | Song3 (blue) | 
|   |   | ...  | 

Column formatting would look nice as added bonus.

And maybe :
library_sorting ; asc | desc in the config that could be used in the `sorted()` functions...

[...]

PR #167, opened 2024-12-02
[WIP] refactor: move mpd client off of ui thread

mierak:
Should mostly solve https://github.com/mierak/rmpc/issues/160


FLchs:
Can also probably help with #147 if querying albums information on each album is implemented.

mierak:
> Can also probably help with #147 if querying albums information on each album is implemented.

Possibly, it could also help with the issues discussed here https://github.com/mierak/rmpc/pull/121. But it wont get us away from having to download the data.

PR #171, opened 2024-12-07
fix: adjust scrollbar position in browser panes

soifou:
In browser panes, the scrollbar position overlaps the pane separator.
This fits nicely with defaults symbols.

However, for a more minimalist approach with empty symbols, eg.
`scrollbar: (symbols: ["", "│", "", ""])`, the scrollbar becomes
invisible or bearly visible depending on the chosen character.

Assume that if the `track` symbol is empty, we expect the scrollbar
to be visible but not overlap with the pane separator.

Solution: Shift the horizontal position to the left.

This ensures the scrollbar remains visible and properly positioned
in all cases, matching the queue appareance.


mierak:
oh yeah, that makes sense, thanks!

PR #172, opened 2024-12-07
fixup: handle scrollbar position when no border

soifou:
The position need to be adjusted as well when `draw_borders` is false.


soifou:
Sorry I should have tested all the cases!

mierak:
all is fine, i should have realized too

PR #173, opened 2024-12-12
feat: show album date in (album)artists panes

mierak:
fixes https://github.com/mierak/rmpc/issues/147


soifou:
Nice feature! But seems to break `Add a song to queue` (adding an album is ok)

mierak:
It also broke environment for external commands because the browser panes rely on the displayed names and this completely borked it. Maybe I should rework it so the items carry additional info instead of just the name..., but this should hopefully fix it for now.

Issue #178, opened 2024-12-16
Ability to display current song at the top/middle of queue panel on startup


MayRedwood:
### Description

If the MPD queue is long enough to exceed the number of lines available for its panel, then, by default, `rmpc` will scroll the panel just enough so that the current song lies at the bottom. This is inconvenient - I find that the quickest way to take a glance at the next few songs in my MPD queue is opening up `ncmpcpp`, as it scrolls the panel enough so that the current song lies at the middle.

Screenshot from `rmpc`:
![image](https://github.com/user-attachments/assets/9b3e5f72-61f3-490c-adc2-7909718ed5f5)

Screenshot from `ncmpcpp`:
![image](https://github.com/user-attachments/assets/b7ade8a5-41a6-437c-9d05-2b2b03a0c260)

### Proposed solution

Being able to configure the current song to show up in the middle or top of the queue panel (or having that be the default behavior, honestly) would be nice.

mierak:
There is a [scrolloff](https://mierak.github.io/rmpc/configuration/#scrolloff) option which applies to the queue table as well as browser panes. It should apply even when song changes and the queue is not in focus.
One minor issue is that it currently does not apply on the very first render in queue, thats on my todo list somewhere.

Does that option do what you want? I dont mind giving more priority to the first render thing either.

MayRedwood:
That does work, yeah! Unfortunately, even with that, the current song is _still_ at the bottom when I open the app. After an up/down input, though, it works as expected. I'm pretty sure this is a bug.

Immediately upon opening:
![image](https://github.com/user-attachments/assets/c0bde826-aab0-4389-b957-71c9e76ffaa2)

After pressing `j` once:
![image](https://github.com/user-attachments/assets/91b1ee71-990a-4186-b8dc-77db8ab9735b)

mierak:
Yup, that is the first render thing I was talking about. It is because rmpc does not know the size of the viewport *before* the first render.
It is indeed a bug and will get fixed soon ish.

mierak:
This has been fixed in https://github.com/mierak/rmpc/pull/180. Scrolloff is now applied on startup in queue.

PR #180, opened 2024-12-19
feat: apply scrollof in queue on start

mierak:
fixes https://github.com/mierak/rmpc/issues/178


soifou:
Tiny feedback, the display of the cover seems broken with this commit

image

mierak:
Hmm, seems like it breaks with `border_type: Full`, works fine with `None`, thats why I didnt notice. Will look into that, thanks again for noticing my mistakes!

mierak:
Should be fixed in https://github.com/mierak/rmpc/pull/181. Dumb mistake, I was rendering the borders into the pane's content area.

PR #189, opened 2024-12-30
refactor: image protocols

mierak:
This greatly simplifies and unifies the code around the different image protocol backends and gets it to a place where I am generally happy with it. I was able to remove a bunch of hacks and improper solutions while also allowing better handling of resize events, simplifying the program flow and other things.
One unfortunate side effect is that Kitty's image is no longer centered anymore and is instead rendered in the top left corner, but this at least brings it in line with the other backends and I plan to introduce image alignment config at some point this release anyway.

Anyway, this is not 100% ready yet but is almost there. Its mostly testing whether something is broken thats left, so if you stumble upon this before merge and feel like testing it out, feel free to do so, thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,13 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Album art with kitty image protocol is no longer centered and is in line with the other image backends
+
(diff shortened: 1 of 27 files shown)


mierak:
looks good to me now

PR #208, opened 2025-01-16
feat: base layout configuration

mierak:
This PR introduces a new layout option in your theme as well as several new Panes. All the unfocusable panes (the new ones, lyrics and album art) can be put inside your layout which makes them visible in all tabs. 

<details>
<summary>Example: Album art is right next to the header, visible on all panes, progress bar is moved between tabs and the header and Queue table can then take the whole tab content space</summary>

image

This is achived with the following theme config:
```rust
layout: Split(
    direction: Vertical,
    panes: [
        (
            size: "8",
            pane: Split(
                direction: Horizontal,
                panes: [
                    (
                        size: "21",
                        pane: Pane(AlbumArt),
                    ),
                    (
                        size: "100%",
                        pane: Split(
                            direction: Vertical,
                            panes: [
                                (
                                    size: "5",
                                    pane: Pane(Header),
                                ),
                                (
                                    size: "1",
                                    pane: Pane(ProgressBar),
                                ),
                                (
                                    size: "3",
                                    pane: Pane(Tabs),
                                ),
                            ]
                        )
                    ),
                ]
            ),
        ),
        (
            size: "100%",
            pane: Pane(TabContent),
        ),
    ],
),

```
</details>

By default it should look and behave the same as it did until now. This was quite a big change to the core UI code so I expect some stuff to break, but hopefully only minor stuff.

To be done:
- [ ] Docs
- [ ] Some more testing

If you want to help out with testing this, please do!


abiud254:
How do the sizes work?

mierak:
I don't notice any more issues in my normal usage. Hopefully all is fine, please open an issue if you find something not working.

Issue #230, opened 2025-01-27
Automatically fetch lrc files


wvffle:
### Description

I have a big catalog of music but I don't have any lyrics. Automatic fetching if the lrc file is absent would be awesome.

### Proposed solution

If the lrc file is absent, automatically fetch it from https://lrclib.net

mierak:
You can use [on_song_change](https://mierak.github.io/rmpc/next/guides/on_song_change/) to automatically fetch lyrics with a bit of scripting. I don't think that hard coding dependency on third-party services into rmpc is the way to go as there are many of them and none are satisfactory enough for general usage.

One downside to this approach is that it wont work for the song that is currently playing since the script has no way to notify rmpc that the download finished. But the next time you play that songs the lyrics will already be there. I plan on improving this in the future but currently have no thought out solution in mind.

wvffle:
That's the way I did it yesterday. However, I have a remote mpd server where the directory names aren't exactly like `artist/album/track-name` and it seems like playing the track for the second time does not read the lrc file.

I've added the `ar`, `al` and `ti` tags, so it could be reindexed when I reopen rmpc.

```bash
      LRCLIB_INSTANCE="https://lrclib.net"

      SONG_OUTPUT="$(rmpc song)"
      ALBUM_NAME="$(echo "$SONG_OUTPUT" | jq -r '.metadata.album')"
      TRACK_NAME="$(echo "$SONG_OUTPUT" | jq -r '.metadata.title')"
      ARTIST_NAME="$(echo "$SONG_OUTPUT" | jq -r '.metadata.artist')"

      LRC_DIR="~/Music/lyrics/$ARTIST_NAME/$ALBUM_NAME/"
      LRC_FILE="$LRC_DIR/$TRACK_NAME.lrc"

      if [ -f "$LRC_FILE" ]; then
          exit 0
      fi

      mkdir -p "$LRC_DIR"

      ENDPOINT="$LRCLIB_INSTANCE/api/get?artist_name=$(echo $ARTIST_NAME | jq -rR @uri)&track_name=$(echo $TRACK_NAME | jq -rR @uri)&album_name=$(echo $ALBUM_NAME | jq -rR @uri)"
      LYRICS="$(curl "$ENDPOINT" | jq -r '.syncedLyrics')"

      if [ "$LYRICS" != "null" ]; then
          echo "[ar:$ARTIST_NAME]" > "$LRC_FILE"
[...]

PR #232, opened 2025-01-29
[WIP] feat: auto fetch lrc

mierak:
This PR introduces remote communication to rmpc which allows scripts and cli to interact with running rmpc instance via unix socket. This can currently be used to display notification messages in rmpc from scripts and to notify rmpc about lyrics being downloaded.

The remote communication protocol, while extremely simplistic, is "private" and not documented by design so we can change it in the future if we need to, it is instead wrapped inside rmpc's CLI.

Fixes https://github.com/mierak/rmpc/issues/230

### DIFF
--- a/docs/src/content/docs/next/guides/on_song_change.mdx
+++ b/docs/src/content/docs/next/guides/on_song_change.mdx
@@ -10,12 +10,20 @@ import { path } from "../data.ts";
 
 Rmpc provides <a href={path("configuration#on_song_change")}>on_song_change</a> property in the config file
 which can be used to run a command whenever the song changes. This can be used for various purposes
-like showing a desktop notification.
+like showing a desktop notification, automatic lyrics download and more.
+
+All song metadata are available to the script as environment variables. For example `$TITLE`, `$FILE`, `$DURATION`, etc.
+
+Additional environment variables available to the script are:
+
+- `$VERSION` - rmpc's version
+- `$PID` - intended for use with the `remote` command and allows it to target specific rmpc instance
+- `$LRC_FILE` - contains the path where rmpc expects to find the lrc file
+- `$HAS_LRC` - indicates whether any lyrics were found for the current song.
 
 ## Desktop notification on song change
 
 Assuming you have a notification daemon with support for images like [dunst](https://dunst-project.org/) running.
-All song metadata are available to the script as environment variables. For example `$TITLE`, `$FILE`, `$DURATION`, etc.
 
 <Steps>
 1. Create a script and place it somewhere. For example `~/.config/rmpc/notify`. Below is an example of such script.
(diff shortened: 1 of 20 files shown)


docs/src/content/docs/next/guides/on_song_change.mdx
```diff
@@ -130,3 +130,43 @@ for the script to execute and increment the play count.
...
+Provided by [wvffle on github](https://github.com/mierak/rmpc/issues/230)
+
+```bash
+#!/usr/bin/env bash
+
+if [ "$HAS_LRC" = "false" ]; then
+    LRCLIB_INSTANCE="https://lrclib.net"
+    LYRICS_DIR="$HOME/Music/lyrics"
+
+    LRC_FILE="$LYRICS_DIR/$ARTIST/$ALBUM/$LRC_DIR/$TITLE.lrc"
```
wvffle, line 145: The docs indicate that:

> Lrc files are resolved by rmpc via two methods(in order):
>   - Same path as the song file, except with the .lrc file extension

Maybe it's worth passing a `LRC_PATH` env var that is an exact path that rmpc expects the lrc file to be in?

That could simplify the script and make it more generic. We could then drop the injection of the `ar`/`al`/`ti` tags as it seems extremely flaky to me.

docs/src/content/docs/next/guides/on_song_change.mdx
```diff
@@ -130,3 +130,43 @@ for the script to execute and increment the play count.
...
+## Automatically fetch lyrics
+
+Provided by [wvffle on github](https://github.com/mierak/rmpc/issues/230)
+
+```bash
+#!/usr/bin/env bash
+
+if [ "$HAS_LRC" = "false" ]; then
+    LRCLIB_INSTANCE="https://lrclib.net"
+    LYRICS_DIR="$HOME/Music/lyrics"
```
wvffle, line 143: I'd move those two above the condition, so that people could see those as configuration vars

docs/src/content/docs/next/guides/on_song_change.mdx
```diff
@@ -130,3 +130,43 @@ for the script to execute and increment the play count.
...
+    LYRICS="$(curl -X GET -sG \
+        --data-urlencode "artist_name=$ARTIST" \
+        --data-urlencode "track_name=$TITLE" \
+        --data-urlencode "album_name=$ALBUM" \
+        "$LRCLIB_INSTANCE/api/get" | jq -r '.syncedLyrics')"
+
+    if [ -z "$LYRICS" ] || [ "$LYRICS" == "null" ]; then
+        echo "Lyrics not found"
+        exit 1
+    fi
```
wvffle, line 161: Thechnically empty value would be when there was a curl error (ie. no internet connection) and not when we 404ed

docs/src/content/docs/next/guides/on_song_change.mdx
mierak: Its already technically possible because you have `$FILE` available inside the script and you should know the lyrics dir. But I dont mind exposing the whole path.

The tags should not really be flaky unless you change the metadata of your library, but there is no harm in having both. Rmpc should prefer matching by path.

mierak: `$LRC_FILE` is now exposed to the script by rmpc, the script has been updated to to reflect the changes

docs/src/content/docs/next/guides/on_song_change.mdx
mierak: I have split the this condition into two and they each now display a different status message in rmpc via remote command

PR #239, opened 2025-02-04
[WIP] feat: better borders in tabs/layout

mierak:
This PR deprecates the `border_type` property on tabs, it will be silently ignored. Instead you can now configure the borders per Pane/Split which allows you to replicate the former behavior while providing more control over them at the same time.

The borders are specified as `borders: "<value>"` where value can be `NONE`, `ALL`, `RIGHT`, `LEFT`, `TOP`, `BOTTOM`, their combination, ie. `borders: "RIGHT | LEFT"` or omitted altogether which implies `NONE`.

Note that you now have to account for the borders if you use exact sizing of panes. For example if your `Header` is supposed to be 2 rows tall and has borders at `TOP` and `BOTTOM` you must set its size as `4` as seen in the example below.

Consider this layout:  

<details>

<summary>Click to expand</summary>

```rust
layout: Split(
    direction: Vertical,
    panes: [
        (
            borders: "ALL",
            pane: Pane(Header),
            size: "4",
        ),
        (
            pane: Pane(Tabs),
            size: "3",
        ),
        (
            borders: "ALL",
            pane: Pane(TabContent),
            size: "100%",
        ),
        (
            size: "3",
            pane: Split(
                direction: Horizontal,
                panes: [
                    (
                        borders: "ALL",
                        pane: Pane(ProgressBar),
                        size: "100%",
                    ),
                ]
            ),
        ),
    ],
),
```

</details>

Along with this tab definition:

<details>

<summary>Click to expand</summary>

```rust
(
    name: "Queue",
    pane: Split(
        direction: Horizontal,
        panes: [
            (
                size: "40%",
                borders: "RIGHT",
                pane: Pane(AlbumArt),
            ),
            (
                size: "60%",
                pane: Pane(Queue),
            ),
        ],
    ),
),

```

</details>

Produces this result:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 - Introduced `StateV2` property as a replacement for `State`. It has additional config properties compared to its predecesor.
 - Introduced `RandomV2`, `ConsumeV2`, `RepeatV2` and `SingleV2` properties as a replacement for their respective earlier versions.
 They have additional config properties compared to their predecesors.
+- `borders` configuration in the tabs configuration
 
 ### Changed
 
@@ -51,6 +52,7 @@ They have additional config properties compared to their predecesors.
 
 ### Deprecated
 
+- `border_type` in tabs config. It has been replaced by the new and more powerful `borders`
 - `theme.tab_bar.enabled` replaced by layout configuration
 - `State` header property
 - `RandomV2`, `ConsumeV2`, `RepeatV2` and `SingleV2` header properties
--- a/docs/src/content/docs/next/assets/example_config.ron
+++ b/docs/src/content/docs/next/assets/example_config.ron
@@ -114,7 +114,6 @@
     tabs: [
         (
             name: "Queue",
-            border_type: None,
             pane: Split(
                 direction: Horizontal,
                 panes: [(size: "40%", pane: Pane(AlbumArt)), (size: "60%", pane: Pane(Queue))],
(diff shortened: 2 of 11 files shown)


mierak:
There are no issues in my testing, please open a new issue if you find some!

soifou:
Nice refactor, I adapted my config using this new layout and, so far, this works perfectly, thanks!

Some thoughts about this subject:

1. I've always found `draw_borders` to be somewhat misleading since it only adds borders to certain parts of the UI — specifically around tabs and the vertical separators in the dirstack. The documentation for this setting is also quite vague. Now that you can achieve more or less the same effect for tabs using borders, the only missing feature is the ability to manage borders for the dirstack. Why not assume that if `TabContent` has full borders, the dirstack should have separators as well? This way, we could deprecate the `draw_borders` setting, or perhaps repurpose it to serve its original semantic purpose: toggling overall borders on or off.
2. Allowing users to manage borders as they like is great, but wouldn’t it be even better to support border styles as well? For example, popups use rounded borders, while `borders: "ALL"` implies solid borders. This creates a mismatch in the overall style.
3. It might be worth mentioning in the doc that tweaking borders requires adjusting sizes accordingly. For instance, combining `borders: ALL` with `size: 2` will only draw borders, hiding the content.

mierak:
> I've always found draw_borders to be somewhat misleading 

I dont like it either and was planning on deprecating it altogether, but I kinda like both your suggestions. I will sit on this one for a bit.

> Allowing users to manage borders as they like is great, but wouldn’t it be even better to support border styles as well? For example, popups use rounded borders, while borders: "ALL" implies solid borders. This creates a mismatch in the overall style.

The biggest issue I have here is that I would like to be able to have connected borders (for example if you create a grid layout with single inner borders, now there are gaps between them). There is an issue opened in ratatui for that to happen automagically https://github.com/ratatui/ratatui/issues/1468 so I am sort of waiting for whats going to happen there, because I don't know how the different border styles will interact. But I definitely want to introduce border style config as well.

> It might be worth mentioning in the doc that tweaking borders requires adjusting sizes accordingly. For instance, combining borders: ALL with size: 2 will only draw borders, hiding the content.

I swear I had it in there :smile:, I will take a look

Thanks for the suggestions as always!

v-Nyo:
> automagically https://github.com/ratatui/ratatui/issues/1468 so I am sort of waiting for whats going to happen there

Got merged: https://github.com/ratatui/ratatui/pull/1874

PR #243, opened 2025-02-06
fix: case-insentive comparison in Property matching

soifou:
When filtering the queue list, capital letters were not accounted for,
leading to no result. With this change, users can now filter without
worrying about exact capitalization.


soifou:
Stumble upon this a couple of time and get today a couple of minutes to look into it!

mierak:
Thanks!

PR #257, opened 2025-02-18
feat(ui): add backdrop support for modals

soifou:
This change enhances the UI by optionally dimming the background when
modals are displayed, improving focus and user experience.

An opt-in setting `modal_backdrop` is introduced in the theme
configuration to enable or disable (default) modal backdrops.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `theme.modal_backdrop` adds a visual backdrop to modals
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


soifou:
Before 

1739901955-region

After

1739901930-region

mierak:
Looks neat, I would put
```rust
        if context.config.theme.modal_backdrop && !self.modals.is_empty() {
            let buffer = frame.buffer_mut();
            buffer.set_style(*buffer.area(), Style::default().fg(Color::DarkGray));
        };
```

directly above https://github.com/mierak/rmpc/blob/4db17dd7c3da32f9c2333bc8d7642acaf1e531c9/src/ui/mod.rs#L174 so you don't have to apply the backdrop in every modal separately

other than that it looks good

soifou:
Ah yes thanks, that's much better, I was too focus on the modal thing that I didn't look back further :)

mierak:
thanks!

Issue #254, opened 2025-02-18
Flaky test `interleaves_repeated_and_scheduled_jobs`


donovanglover:
### Description

The test `core::scheduler::tests::interleaves_repeated_and_scheduled_jobs` was introduced in https://github.com/mierak/rmpc/commit/1da4d89cc3653dd682230df6ec0cfa257e399f77 and seems to occasionally fail in Darwin CI.

- Test fails: https://github.com/NixOS/nixpkgs/pull/382709/checks?check_run_id=37328082702
- Test passes: https://github.com/donovanglover/nixpkgs-review-gha/actions/runs/13387034999/job/37385999748
- Test fails: https://github.com/donovanglover/nixpkgs-review-gha/actions/runs/13387034999/job/37386000101

### Steps to reproduce

Fork https://github.com/Defelo/nixpkgs-review-gha and enter `382709` as the pull request number

### Expected behavior

Not flaky

### rmpc debuginfo

None

### Relevant log output

```rust
failures:
---- core::scheduler::tests::interleaves_repeated_and_scheduled_jobs stdout ----
thread 'core::scheduler::tests::interleaves_repeated_and_scheduled_jobs' panicked at src/core/scheduler/mod.rs:372:9:
assertion `left == right` failed
  left: [5, 6, 1, 2, 7, 8, 9, 1, 2]
 right: [5, 6, 1, 2, 7, 8, 1, 2, 9]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
    core::scheduler::tests::interleaves_repeated_and_scheduled_jobs
test result: FAILED(B. 428 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.51s
```

mierak:
Interesting, I was sort of afraid of this so I ran these tests for quite a while in a loop and it seemed fine on my PC and my CI. I will take a look.

You can skip skip this test for now if its blocking you.

donovanglover:
From the logs:

```rust
cargoCheckHook flags: -j 3 --profile release --target aarch64-apple-darwin --offline -- --test-threads=3
```

Also see [`cargo-check-hook.sh`](https://github.com/NixOS/nixpkgs/blob/d2ab2691c798f6b633be91d74b1626980ddaff30/pkgs/build-support/rust/hooks/cargo-check-hook.sh#L41)

mierak:
[...]

PR #259, opened 2025-02-18
test

mierak:
flaky tests on macos only, see https://github.com/mierak/rmpc/issues/254


soifou:
Did a quick and cheap bench on my macbook pro m2 by running 100 iterations of the test suite under normal and "heavy" load (stress test multiple cores)

Here are the failures using the current durations

|        | Normal load | Heavy load |
|--------|-------------|------------|
| Test 1 | 1           | 36         |
| Test 2 | 0           | 45         |
| Test 3 | 0           | 31         |

After increasing all durations by a factor of 4, I got way better results:

|        | Normal load | Heavy load |
|--------|-------------|------------|
| Test 1 | 0           | 0          |
| Test 2 | 0           | 1          |
| Test 3 | 0           | 0          |

So I throw my 2 cents that this might be due to differences in timing precision between Linux and macOS kernels...
Have you observed something similar when stress testing and adjusting the durations on linux?

mierak:
Yeah, there definitely are differences when I make the times smaller. I might have to rethink these tests completely but I have no idea how to make them time independent, its a shame really since they are consistent on linux.

mierak:
Stabilized by abstracting away dependency on time into a `TimeProvider` trait which is replacable in tests.

PR #267, opened 2025-03-07
feat: on_resize hook

mierak:


soifou:
Nice feature! 

Will you also consider an optional CLI `--theme` argument, eg. `rmpc --theme foo`, that would take precedence over the one set in config, `theme: Some("bar")`? 

This way we could spawn several rmpc instances with different themes while keeping the exact same configuration.

mierak:
There already is --config flag which allows you to do that. But I suppose it does not hurt to add --theme too so you can keep the same config as you mentioned.

PR #268, opened 2025-03-07
feat: theme cli arg

mierak:
@soifou Is it alright like this? It requires an absolute path to the theme, ie `~/.config/rmpc/themes/theme.ron`.

Changing it to be able to just use a theme name is possible but complicates it a bit because we have to check whether we even have config file etc.

edit: this way it is also consistent with the way the IPC theme command is done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - rmpc will now reload config changes automatically
 - remote command to change theme
 - `on_resize` which is called whenever rmpc is resized
+- `--theme` cli argument to override theme in the config file
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


soifou:
I was creating an issue for that, but you're damn fast! Let me try and I get back to you, but yeah this what I was thinking, thanks so much

soifou:
Yes tried and adopted 👍  
Thanks again for the prompt implementation, much appreciated!

soifou:
> It requires an absolute path to the theme, ie ~/.config/rmpc/themes/theme.ron.

Relative path works too, and I'm very ok with that. I wanted this scriptable.

mierak:
Alright, merging, thanks!

PR #271, opened 2025-03-09
feat(lyrics): don't highlight until the first lyrics have been reached

Kladki:
Before, the first lyrics would be highlighted even before it's own timestamp was reached


mierak:
Looks good to me. Thanks!

PR #272, opened 2025-03-09
fix(lyrics): parse lyric timestamps with frations of seconds which aren't to 2s.f.

Kladki:
Ik that it's not mentioned on the Wikipedia page (prob since LRC isn't really standardised), but some lyric providers (Netease and Apple Music for example) use 3s.f. for the fraction part of a second, causing the most significant digit to be treated as a second. This fixes the issue, by letting any number of s.f. to be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Lyrics with fractions of seconds which weren't to 2s.f. being
+  parsed incorrectly
+
 - Album art staying on the old one when in tmux and not visible
 
 ## [0.8.0] - 2025-02-16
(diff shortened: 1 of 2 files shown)


mierak:
Thanks for the PR! Can you also add this to the tests? Does not have to be a separate test case, just adding a lrc line to the first test is fine.

mierak:
Looks good, please get rid of the expect as well and we can merge

mierak:
Replace `expect` with `context` and use `?` operator on the context function call.

its a function from anyhow https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context

mierak:
It would work even without the context because the from_str() and try_from() both return a result already, but this way it allows us to keep the context of what went wrong that you had in the expect.

Anyway, merging, thanks again!

PR #277, opened 2025-03-12
Update documentation to mention `Alt` key`

neio4:
In emacs `Alt` is called `Meta`. I also mentioned that in order to make searching for `meta` possible.


mierak:
thanks!

PR #282, opened 2025-03-14
Add nord-theme to docs

v-Nyo:
Should be alright now?


mierak:
Nord is a well known colorscheme and wont look like nord when you use the therminal colors no? You can also credit yourself if you want.

mierak:
Nord theme defines its color palette, the ones you have in the hex theme so that is the one I would keep and discard the other one if you want to call this Nord. The other one wont look like nord on my system at all for example.

> I made a separate file with only hex codes. 2 Issues I can't figure out how to theme this text:

yeah, that is something I have on the backburner, its the last hardcoded color in rmpc which you cant change at the moment.

> I assume it's https://mierak.github.io/rmpc/release-0-7-0/configuration/theme/#browser_song_format

That describes how to display each song entry in the browser panes, for example `<track number> <artist> - <song>` by default. The styles are ignored in here as it says in the docs.

mierak:
> Do you plan to add opacity to the colors?

Unfortunately there is nothing like transparency in the terminal.

> And do you know what that weird bar is I mentioned?

That is most likely your window manager or maybe the terminal emulator itself

mierak:
> Is this currently also not configurable? 20250315_15h09m06s_grim

Yeah, you're right, this is also not configurable at the moment

> Never mind as far as I can tell kitty gives you transparency when the bg color is the same as in kitty (specified or not), all other colors don't work

Yup, theres no transparency setting for colors in the terminal. But as you have found out, if you set your terminal emulator to be transparent and do not set any background color in rmpc it will be transparent.

PR #288, opened 2025-03-24
feat: property pane background color

mierak:
Currently the property pane was not able to have a proper background different from the normal background color, only background color for the property could be specified, but that does not necessarily fill the whole pane.


mierak:
Scratch that, it will be better to support background colors for all the pane types, not just the property.

PR #292, opened 2025-03-31
Switching to a matrix in the CI

Xithrius:
Hey, great project! Came across this while going through the [console clients for mpd in the arch wiki](https://wiki.archlinux.org/title/Music_Player_Daemon#Console).

Was thinking this might be useful if you want modify expand/modify CI runs for building and testing without much thought. Included ubuntu/macos on stable and nightly for now.

Thanks!


mierak:
Hi, thanks, looks great! Two things though:

* I would like to keep tests and clippy for both debug and release mode, theres some differences in the code between those and I ran into issues because of that before
* Rustfmt needs to be on nightly since it uses unstable features

Xithrius:
Sure thing, I'll modify the CI to have those changes. Thanks!

Xithrius:
There's a lot of formatting fixes across a bunch of files that come with nightly clippy -- I can fix them here or make another PR dedicated to that.

mierak:
Only rustfmt should be on nightly because rmpc uses unstable features in rustfmt.toml. Clippy is currently checking on stable, so thats probably why. Lets keep clippy on stable.

mierak:
great, thanks!

PR #294, opened 2025-03-31
Handle new value lastloadedplaylist from mpd

Xithrius:
Fixes terminal getting spammed with warnings that this key/value pair of `lastloadedplaylist` not being handled:

```
2025-03-31T08:09:45.862124626Z WARN  thread=request src/mpd/mod.rs:21 message="Encountered unknow key/value pair" key="lastloadedplaylist" value=""
```

I understand it's not being used at the moment, so if we want to we could probably just set this to `None` as well, let me know what you think.

Last loaded playlist is a string, which can be seen in [mpd's src/queue/Queue.hxx, tag v0.24.2](https://github.com/MusicPlayerDaemon/MPD/blob/v0.24.2/src/queue/Queue.hxx#L89).

Also went ahead and fixed that typo which existed [here](https://github.com/mierak/rmpc/blob/d5ec479eef0cb34df7b1d7ee5e5d034b7a33aca3/src/mpd/mod.rs#L21) for `FromMpd::next`, `LineHandled::No` case.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "[rust]": {
+    "editor.defaultFormatter": "rust-lang.rust-analyzer"
+  },
+  "rust-analyzer.check.command": "clippy",
+  "rust-analyzer.rustfmt.extraArgs": ["+nightly"]
+}
--- a/src/mpd/commands/status.rs
+++ b/src/mpd/commands/status.rs
@@ -44,6 +44,7 @@ pub struct Status {
                                     * explanation. */
     pub updating_db: Option<u32>, // job id
     pub error: Option<String>,    // if there is an error, returns message here
+    pub lastloadedplaylist: Option<String>, // last loaded stored playlist
 }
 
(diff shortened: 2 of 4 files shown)


mierak:
Either or if fine by me here. Thanks for fixing the typo as well!

The formatting needs to be on nightly toolchain, I suppose its time to create contribution notes somewhere to mention that.
For the clippy warning, feel free to just silence it, thats a problem for different time.

Xithrius:
Added VSCode settings as well so a trigger of formatting (from autosave or otherwise) will use nightly.

Xithrius:
Hey @mierak, does this all look good? Thanks!

mierak:
There is a failing clippy lint that needs to be resolved, other than that its fine.

As I have said here
> For the clippy warning, feel free to just silence it, thats a problem for different time.

You can just silence it so the build does not fail.

Xithrius:
My bad, missed the failing lint. Should be good now.

Issue #263, opened 2025-03-02
Scripting/templating for strings


roxwize:
### Description

Maybe this is just my past experience with foobar2000 rotting my brain into wanting this sort of thing from other players, but I think it'd be a nice improvement if this:

```rust
(kind: Text("[RMPC] ")),
(kind: Property(Song(Artist))),
(kind: Text(" - ")),
(kind: Property(Song(Album))),
(kind: Text(" - ")),
(kind: Property(Song(Title)))
```

could possibly be made more intuitive to write out by instead having it be represented as a string that takes in variables, like:

```
[RMPC] $artist - $album - $title
```

Of course, it wouldn't necessarily need to be a full-fledged scripting system with built-in utility functions and such like what foobar has, but it still makes writing things that output an informative string much easier to configure.

### Proposed solution

Naturally ideas can be borrowed from what foobar does: variables represented as `%name%`, functions as `$name(args)`, and defaults as `[ ]`. I think a good minimal approach that should cover most bases would be to only do the first and last parts. Variables referring to song information &c., and brackets to handle when information has the potential to not exist. An example: if I were to specify `%artist%[ - %title%]`, then `Artist` would display if the track is untitled; otherwise, it will display the full `Artist - Title`.

I'm in favor of representing variables as simply `%var`; the additional percentage at the end appears clunky and useless since variables shouldn't contain any spaces anyways. Brackets could be escaped with `\[` and `\]` respectively.

An example of a "function" that could be implemented in the templating/scripting language, if feasible, would be a function to pad strings with spaces or other characters. In music players like foobar I would always display the track number of the music I listen to with padded zeroes so that everything would line up all the time, especially with my monospace font. So, something like turning 5 into 05 could be achieved as such: `$pad($track_number, 2, "0")`

Again, this isn't something that would need to be fully-featured. A system in place to just easily specify sequences of string variables without having to use an entire array with a lot of enums would make configuration much easier. I'm no expert in Rust but I could probably give this a try if it isn't deemed out of scope for this software.

mierak:
So this is something I wanted to do from the start but could not settle on syntax. Though I have/had no ambitions for functions as they are not possible even through the current config.

Padding for track number is currently hardcoded to two numbers since I had that requirement as well. https://github.com/mierak/rmpc/blob/65b5d949435a2f34880355526dc3595d5f27e85a/src/ui/panes/mod.rs#L575

[...]

PR #309, opened 2025-04-22
[WIP] feat: property template string

mierak:
This is a working implementation of templating as discussed here https://github.com/mierak/rmpc/issues/263. I am mostly looking at opinions/suggestions at this point. Tagging you because you have shown interest in this @roxwize.

Currently it works only for `Property` panes in the layout and you have to provide `format` instead of `content`.

It allows to transform for example this: 
```ron
(
    size: "23",
    pane: Pane(Property(
        content: [
            (kind: Property(Status(Elapsed))), 
            (kind: Text(" / ")), 
            (kind: Property(Status(Duration))), 
            (kind: Group([
                (kind: Text(" (")), 
                (kind: Property(Status(Bitrate))), 
                (kind: Text(" kbps)")),
            ])),
        ], 
        align: Left,
    )),
),
```

into this:
```ron
(
    size: "23",
    pane: Pane(Property(
        format: "$elapsed$' / '$duration$[$' ('$bitrate$' kpbs)']",
        align: Left,
    )),
),
```

These definitions are equivalent.

Groups are 

Some syntax remarks:
* `Group` kind are specified as a dollar sign and square brackets`$[ ... ]`
* `Text` kind is a dollar sign and either single or double quotes: `$'this is raw text'`, `$"also raw text"`, `$"escaping \"works too\""`
* `Sticker` is currently `$sticker(name: "playCount")`'
* other properties are similar: `$consume(onLabel: "on", offLabel: "off")`
* `default`(fallback value) is delimited by `|`: `$title|$'No song'`
* styles are in square braces following a property: `$title{fg: yellow, bg: black, mods: bold}|$'No song'{fg: red, bg: black}`

You can see some tests in parser.rs to see more examples.

### DIFF
--- a/src/config/tabs.rs
+++ b/src/config/tabs.rs
@@ -1,7 +1,8 @@
 #![allow(deprecated)] // TODO remove after cleanup
 use std::collections::HashMap;
 
-use anyhow::{Result, ensure};
+use anyhow::{Result, bail, ensure};
+use chumsky::Parser;
 use derive_more::{Deref, Display, Into};
 use itertools::Itertools;
 use ratatui::{layout::Direction, widgets::Borders};
@@ -9,6 +10,7 @@ use serde::{Deserialize, Serialize};
 
 use super::theme::{
     PercentOrLength,
+    parser,
     properties::{Property, PropertyFile, PropertyKind, PropertyKindFile},
 };
 use crate::shared::id::{self, Id};
@@ -53,7 +55,10 @@ pub enum PaneTypeFile {
     #[cfg(debug_assertions)]
     FrameCount,
     Property {
-        content: Vec<PropertyFile<PropertyKindFile>>,
+        #[serde(default)]
+        content: Option<Vec<PropertyFile<PropertyKindFile>>>,
+        #[serde(default)]
+        format: Option<String>,
         #[serde(default)]
         align: super::theme::properties::Alignment,
         #[serde(default)]
@@ -128,9 +133,11 @@ impl Pane {
     }
 }
 
-impl From<PaneTypeFile> for PaneType {
-    fn from(value: PaneTypeFile) -> Self {
-        match value {
+impl TryFrom<PaneTypeFile> for PaneType {
+    type Error = anyhow::Error;
+
+    fn try_from(value: PaneTypeFile) -> std::result::Result<PaneType, anyhow::Error> {
+        Ok(match value {
             PaneTypeFile::Queue => PaneType::Queue,
             #[cfg(debug_assertions)]
             PaneTypeFile::Logs => PaneType::Logs,
@@ -148,20 +155,34 @@ impl From<PaneTypeFile> for PaneType {
             PaneTypeFile::TabContent => PaneType::TabContent,
             #[cfg(debug_assertions)]
             PaneTypeFile::FrameCount => PaneType::FrameCount,
-            PaneTypeFile::Property { content: properties, align, scroll_speed } => {
-                PaneType::Property {
-                    content: properties
-                        .into_iter()
-                        .map(|prop| prop.try_into().expect(""))
-                        .collect_vec(),
-                    align: align.into(),
-                    scroll_speed,
-                }
-            }
+            PaneTypeFile::Property { content, align, scroll_speed, format } => PaneType::Property {
+                content: {
+                    match (format, content) {
+                        (Some(format), Some(_)) | (Some(format), None) => parser::parser()
+                            .parse(&format)
+                            .into_result()
+                            .map_err(|e| {
+                                anyhow::anyhow!("Failed to parse property format: {:?}", e)
+                            })?
+                            .into_iter()
+                            .map(|prop| -> Result<_> { prop.try_into() })
+                            .try_collect()?,
+                        (None, Some(content)) => content
+                            .into_iter()
+                            .map(|prop| -> Result<_> { prop.try_into() })
+                            .try_collect()?,
+                        (None, None) => {
+                            bail!("One of config or format has to be specified for a Property pane")
+                        }
+                    }
+                },
+                align: align.into(),
+                scroll_speed,
+            },
             PaneTypeFile::Browser { root_tag: tag, separator } => {
                 PaneType::Browser { root_tag: tag, separator }
             }
-        }
+        })
     }
 }
 
@@ -342,7 +363,7 @@ impl PaneOrSplitFile {
     pub fn convert_recursive(&self, b: Borders) -> Result<SizedPaneOrSplit> {
         Ok(match self {
             PaneOrSplitFile::Pane(pane_type_file) => SizedPaneOrSplit::Pane(Pane {
-                pane: pane_type_file.clone().into(),
+                pane: pane_type_file.clone().try_into()?,
                 borders: b,
                 id: id::new(),
             }),
--- a/src/config/theme/parser.rs
+++ b/src/config/theme/parser.rs
@@ -1,0 +1,661 @@
+use std::collections::HashMap;
+
+use chumsky::prelude::*;
+
+use super::{
+    StyleFile,
+    properties::{
+        PropertyFile,
+        PropertyKindFile,
+        PropertyKindFileOrText,
+        SongPropertyFile,
+        StatusPropertyFile,
+        WidgetPropertyFile,
+    },
+    style::Modifiers,
+};
+
+pub fn parser<'a>()
+-> impl Parser<'a, &'a str, Vec<PropertyFile<PropertyKindFile>>, extra::Err<Rich<'a, char>>> {
+    let ident = text::ascii::ident();
+
+    let escape_double_quoted_str = just('\\')
+        .then(choice((
+            just('\\'),
+            just('/'),
+            just('"'),
+            just('b').to('\x08'),
+            just('f').to('\x0C'),
+            just('n').to('\n'),
+            just('r').to('\r'),
+            just('t').to('\t'),
+            just('u').ignore_then(text::digits(16).exactly(4).to_slice().validate(
+                |digits, e, emitter| {
+                    char::from_u32(
+                        u32::from_str_radix(digits, 16)
+                            .expect("Only valid digits should have been parsed in text::digits"),
+                    )
+                    .unwrap_or_else(|| {
+                        emitter.emit(Rich::custom(e.span(), "invalid unicode character"));
+                        '\u{FFFD}'
+                    })
+                },
+            )),
+        )))
+        .ignored()
+        .boxed();
+
+    let escape_single_queoted_str =
+        choice((
+            just('\\').then(just('\'')).to('a').ignored(),
+            just('\\')
+                .then(choice((
+                    just('\\'),
+                    just('/'),
+                    just('b').to('\x08'),
+                    just('f').to('\x0C'),
+                    just('n').to('\n'),
+                    just('r').to('\r'),
+                    just('t').to('\t'),
+                    just('u').ignore_then(text::digits(16).exactly(4).to_slice().validate(
+                        |digits, e, emitter| {
+                            char::from_u32(u32::from_str_radix(digits, 16).expect(
+                                "Only valid digits should have been parsed in text::digits",
+                            ))
+                            .unwrap_or_else(|| {
+                                emitter.emit(Rich::custom(e.span(), "invalid unicode character"));
+                                '\u{FFFD}'
+                            })
+                        },
+                    )),
+                )))
+                .ignored(),
+        ))
+        .boxed();
+
+    let double_quoted_string = none_of("\\\"")
+        .ignored()
+        .or(escape_double_quoted_str)
+        .repeated()
+        .to_slice()
+        .map(ToString::to_string)
+        .delimited_by(just('"'), just('"'))
+        .boxed();
+
+    let single_quoted_string = none_of("\\'")
+        .ignored()
+        .or(escape_single_queoted_str)
+        .repeated()
+        .to_slice()
+        .map(ToString::to_string)
+        .delimited_by(just("'"), just("'"))
+        .boxed();
+
+    let string = double_quoted_string.clone().or(single_quoted_string.clone()).labelled("string");
+
+    let modifiers = choice((
+        just("bold"),
+        just("dim"),
+        just("italic"),
+        just("underlined"),
+        just("reversed"),
+        just("crossedout"),
+    ))
+    .separated_by(just(',').padded())
+    .collect::<Vec<_>>()
+    .map(|val| {
+        if val.is_empty() {
+            return None;
+        }
+
+        let mut res = Modifiers::empty();
+        for modifier in val {
+            res = match modifier {
+                "bold" => res.union(Modifiers::Bold),
+                "dim" => res.union(Modifiers::Dim),
+                "italic" => res.union(Modifiers::Italic),
+                "underlined" => res.union(Modifiers::Underlined),
+                "reversed" => res.union(Modifiers::Reversed),
+                "crossedout" => res.union(Modifiers::CrossedOut),
+                _ => res,
+            };
+        }
+
+        Some(res)
+    });
+
+    let color = choice((
+        ident.map(|c: &str| StringOrModifiers::String(c.to_owned())),
+        just('#').then(ident).map(|(_, hex): (_, &str)| {
+            let mut hex = hex.to_owned();
+            hex.insert(0, '#');
+            StringOrModifiers::String(hex)
+        }),
+        text::int(10).map(|n: &str| StringOrModifiers::String(n.to_owned())),
+    ));
+
+    let style_file = choice((
+        just("fg").then_ignore(just(':').padded()).then(color),
+        just("bg").then_ignore(just(':').padded()).then(color),
+        just("mods")
+            .then_ignore(just(':').padded())
+            .then(modifiers.map(StringOrModifiers::Modifiers)),
+    ))
+    .separated_by(just(',').padded())
+    .collect::<HashMap<_, _>>()
+    .map(|mut m| StyleFile {
+        fg: m.remove("fg").get_string(),
+        bg: m.remove("bg").get_string(),
+        modifiers: m.remove("mods").get_modifiers(),
+    })
+    .delimited_by(just('{'), just('}'));
+
+    let label = string.clone().map(|v: String| StyleOrLabel::Label(v));
+    let style = style_file.map(StyleOrLabel::Style);
+
+    let status_property = choice((
+        just("volume").map(|_| PropertyKindFile::Status(StatusPropertyFile::Volume)),
+        just("repeat")
+            .ignored()
+            .then(
+                choice((
+                    just("onStyle").then_ignore(just(':').padded()).then(style),
+                    just("offStyle").then_ignore(just(':').padded()).then(style),
+                    just("onLabel").then_ignore(just(':').padded()).then(label.clone()),
+                    just("offLabel").then_ignore(just(':').padded()).then(label.clone()),
+                ))
+                .separated_by(just(',').padded())
+                .collect::<HashMap<_, _>>()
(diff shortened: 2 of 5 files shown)


mierak:
fwiw, I have replaced 1:1 every single property occurence in my theme (close to the default one) and ran into no issues

<details>
<summary>Click to see diff</summary>

```diff
diff --git a/dot_config/rmpc/themes/def.ron b/dot_config/rmpc/themes/def.ron
index a36983f..355c22e 100644
--- a/dot_config/rmpc/themes/def.ron
+++ b/dot_config/rmpc/themes/def.ron
@@ -28,31 +28,16 @@
                                 panes: [
                                     (
                                         size: "23",
-                                        pane: Pane(Property(
-                                            content: [
-                                                (kind: Text("["), style: (fg: "yellow", modifiers: "Bold")),
-                                                (kind: Property(Status(StateV2())), style: (fg: "yellow", modifiers: "Bold")),
-                                                (kind: Text("]"), style: (fg: "yellow", modifiers: "Bold")),
-                                            ], align: Left,
-                                        )),
+                                        pane: Pane(Property(format: "$'['{fg: yellow, mods: bold}$state(){fg: yellow, mods: bold}$']'{fg: yellow, mods: bold}", align: Left)),
                                     ),
                                     (
                                         size: "100%",
                                         borders: "LEFT | RIGHT",
-                                        pane: Pane(Property(
-                                            content: [
-                                                (kind: Property(Song(Filename)), style: (modifiers: "Bold"),
-                                                    default: (kind: Text("No Song"), style: (modifiers: "Bold")))
-                                            ], align: Center, scroll_speed: 1
-                                        )),
+                                        pane: Pane(Property(format: "$s:title{mods: bold}|$'No Song'{mods: bold}", align: Center, scroll_speed: 1)),
                                     ),
                                     (
                                         size: "23",
-                                        pane: Pane(Property(
-                                            content: [
-                                                (kind: Property(Widget(Volume)), style: (fg: "blue", modifiers: "Bold"))
-                                            ], align: Right
-                                        )),
+                                        pane: Pane(Property( format: "$w:volume{fg: blue, mods: bold}", align: Right)),
                                     ),
                                 ],
                             ),
@@ -65,63 +50,22 @@
                                 panes: [
                                     (
                                         size: "23",
-                                        pane: Pane(Property(
-                                            content: [
-                                                (kind: Property(Status(Elapsed))), 
-                                                (kind: Text(" / ")), 
-                                                (kind: Property(Status(Duration))), 
-                                                (kind: Group([
-                                                    (kind: Text(" (")), 
-                                                    (kind: Property(Status(Bitrate))), 
-                                                    (kind: Text(" kbps)")),
-                                                ])),
-                                            ], align: Left,
-                                        )),
+                                        pane: Pane(Property(format: "$elapsed$' / '$duration$[$' ('$bitrate$' kpbs)']", align: Left)),
                                     ),
                                     (
                                         size: "100%",
                                         borders: "LEFT | RIGHT",
-                                        pane: Pane(Property(
-                                            content: [
-                                                (kind: Property(Song(Artist)), style: (fg: "yellow", modifiers: "Bold"),
-                                                    default: (kind: Text("Unknown"), style: (fg: "yellow", modifiers: "Bold"))),
-                                                (kind: Text(" - ")),
-                                                (kind: Property(Song(Album)), default: (kind: Text("Unknown Album")))
-                                            ], align: Center, scroll_speed: 2
-                                        )),
+                                        pane: Pane(Property(format: "$s:artist{fg: yellow, mods: bold}|$'Unknown Artist'{fg: yellow, mods: bold}$' - '$s:album{fg: yellow, mods: bold}|$'Unknown Album'", align: Center, scroll_speed: 2)),
                                     ),
                                     (
                                         size: "23",
-                                        pane: Pane(Property(content: [
-                                            (kind: Property(Status(RepeatV2(
-                                                on_label: " ",
-                                                off_label: " ",
-                                                on_style: (fg: "yellow", modifiers: "Bold"),
-                                                off_style: (fg: "blue", modifiers: "Dim"),
-                                            )))),
-                                            (kind: Property(Status(RandomV2(
-                                                on_label: " ",
-                                                off_label: " ",
-                                                on_style: (fg: "yellow", modifiers: "Bold"),
-                                                off_style: (fg: "blue", modifiers: "Dim"),
-                                            )))),
-                                            (kind: Property(Status(SingleV2(
-                                                on_label: "󰑘 ",
-                                                off_label: "󰑘 ",
-                                                oneshot_label: "󰑘 ",
-                                                on_style: (fg: "yellow", modifiers: "Bold"),
-                                                off_style: (fg: "blue", modifiers: "Dim"),
-                                                oneshot_style: (fg: "red", modifiers: "Bold"),
-                                            )))),
-                                            (kind: Property(Status(ConsumeV2(
-                                                on_label: " ",
-                                                off_label: " ",
-                                                oneshot_label: "",
-                                                on_style: (fg: "yellow", modifiers: "Bold"),
-                                                off_style: (fg: "blue", modifiers: "Dim"),
-                                                oneshot_style: (fg: "red", modifiers: "Dim"),
-                                            )))),
-                                        ], align: Right)),
+                                        pane: Pane(Property(
+                                            format: "
+                                                $repeat(onLabel: ' ', offLabel: ' ', onStyle:  {fg: yellow, mods: bold}, offStyle: {fg: yellow, mods: dim})
+                                                $random(onLabel: ' ', offLabel: ' ', onStyle: {fg: yellow, mods: bold}, offStyle: {fg: yellow, mods: dim})
+                                                $single(onLabel: '󰑘 ', offLabel: '󰑘 ', oneshotLabel: '󰑘 ', onStyle: {fg: yellow, mods: bold}, offStyle: {fg: yellow, mods: dim}, oneshotStyle: {fg: red, mods: bold})
+                                                $consume(onLabel: ' ', offLabel: ' ', oneshotLabel: ' ', onStyle: {fg: yellow, mods: bold}, offStyle: {fg: yellow, mods: dim}, oneshotStyle: {fg: red, mods: dim})
+                                            ", align: Right)),
                                     ),
                                 ],
                             ),
@@ -147,9 +91,9 @@
                     direction: Horizontal,
                     panes: [
                         (
-                            pane: Pane(Property(content: [(kind: Property(Status(StateV2(playing_label: "  ", paused_label: "  ", stopped_label: "  ",
-                                playing_style: (fg: "blue"), paused_style: (fg: "green"), stopped_style: (fg: "red")
-                            ))))], align: Left)),
+                            pane: Pane(Property(
+                                format: "$state(playingLabel: '  ', pausedLabel: '  ', stoppedLabel: '  ', playingStyle: {fg: blue}, pausedStyle: {fg: green}, stoppedStyle: {fg: red})",
+                                align: Left)),
                             size: "3",
                         ),
                         (
@@ -158,13 +102,7 @@
                         ),
                         (
                             size: "13",
-                            pane: Pane(Property(
-                                content: [
-                                    (kind: Property(Status(Elapsed))),
-                                    (kind: Text(" / ")),
-                                    (kind: Property(Status(Duration))),
-                                ], align: Right,
-                            )),
+                            pane: Pane(Property(format: "$elapsed$' / '$duration", align: Right)),
                         ),
                     ]
                 ),
@@ -190,22 +128,7 @@
         thumb_style: (fg: "blue"),
     ),
     browser_column_widths: [20, 38, 42],
-    browser_song_format: [
-        (
-            kind: Group([
-                (kind: Property(Track)),
-                (kind: Text(" ")),
-            ])
-        ),
-        (
-            kind: Group([
-                (kind: Property(Artist)),
-                (kind: Text(" - ")),
-                (kind: Property(Title)),
-            ]),
-            default: (kind: Property(Filename))
-        ),
-    ],
+    browser_song_format_v2: "$[$s:track$' ']$[$s:artist$' - '$s:title]|$s:filename$'.'$s:fileextension",
     tab_bar: (
         active_style: (fg: "black", bg: "blue", modifiers: "Bold"),
         inactive_style: (),
@@ -215,25 +138,10 @@
     borders_style: (fg: "blue", modifiers: "Bold"),
     highlight_border_style: (fg: "red"),
     song_table_format: [
-        (
-            prop: (kind: Property(Other("albumartist")), default: (kind: Property(Artist), default: (kind: Text("Unknown")))),
-            width: "20%",
-            label: "Artist"
-        ),
-        (
-            prop: (kind: Property(Title), default: (kind: Text("Unknown"))),
-            width: "35%",
-        ),
-        (
-            prop: (kind: Property(Album), default: (kind: Text("Unknown Album"))),
-            width: "45%",
-        ),
-        (
-            prop: (kind: Property(Duration),default: (kind: Text("-"))),
-            width: "5",
-            alignment: Right,
-            label: "Len"
-        ),
+        (format: "$s:albumartist|$s:artist|$'unknown'", width: "20%", label: "Artist"),
+        (format: "$s:title|$'Unknown'", width: "35%"),
+        (format: "$s:album|$'Unknown Album'", width: "45%"),
+        (format: "$s:duration|$'-'", width: "5", alignment: Right, label: "Len"),
     ],
     header: (rows: []),
 )
```
</details>

roxwize:
I haven't been able to offer anything to this because I've been dealing with life etc. for a while. Got sick and had some personal issues and stuff so I don't know how long until I'll have anything to offer. I will say that the syntax is a bit strange but I'll get to that when I'm not feeling terrible 24/7.

v-Nyo:
This seems really nice.  The only feedback I can think of right now is maybe taking some inspiration from the waybar config.

Gri-ffin:
@mierak I've been reviewing the PR and the discussions, and this looks really great. That said, I want to challenge a few aspects to explore potential pitfalls, if you don’t mind:

**1. Robust Error Handling and User Feedback**

With this new "mini-language," I think strong error reporting becomes crucial to maintaining a good user experience. While technical users might be able to debug most issues, less technical users, especially once config files become more complex, may struggle to understand vague or cryptic errors.

This might be less of an issue now since the majority of users are on Linux and probably more technical, but I think you should still consider when users who are not much technical might use rmpc when the project become very popular.

**2. Feature Scope and Complexity**

Features like `%padl`, `%if`, and `%progress` are exciting and open the door to powerful customizations. However, I suggest a slow approach here. You should the current core templating syntax first, and then gradually introduce more advanced functions after you gather user feedback.

I believe adding too many features too early might lead to a higher maintenance cost in the long run.

**3. Documentation**

In my eyes, this is the most critical piece moving forward. Since you’re introducing a whole new syntax, a clear and complete documentation is very much essential for adoption.
It’s very very important that users can quickly understand:

- What syntax is valid and what isn’t
- What properties exist and in what contexts they’re allowed
- How groups, default fallbacks, and styles work
- Error examples and how to fix them
- Common patterns and idioms (e.g., fallback chaining, layout composition)

This will make it much easier for users to adopt and use the language and even contribute safely.

src/config/tabs.rs
```diff
@@ -148,20 +155,34 @@ impl From<PaneTypeFile> for PaneType {
...
+                    match (format, content) {
+                        (Some(format), Some(_)) | (Some(format), None) => parser::parser()
+                            .parse(&format)
+                            .into_result()
+                            .map_err(|e| {
+                                anyhow::anyhow!("Failed to parse property format: {:?}", e)
+                            })?
+                            .into_iter()
+                            .map(|prop| -> Result<_> { prop.try_into() })
+                            .try_collect()?,
```
Gri-ffin, line 169: This should be documented, users should know that format will take precedence, or it might spark confusion

src/config/theme/parser.rs
```diff
@@ -0,0 +1,891 @@
...
+                    })
+                },
+            )),
+        )))
+        .ignored()
+        .boxed();
+
+    let escape_single_queoted_str =
+        choice((
+            just('\\').then(just('\'')).to('a').ignored(),
```
Gri-ffin, line 47: Couldn't be just be just('\\').then(just('\'')).to('\'').ignored(), forgive me if i'm wrong, but i don't understand what does the a do here?

src/config/theme/parser.rs
```diff
@@ -0,0 +1,891 @@
...
+            just('f').to('\x0C'),
+            just('n').to('\n'),
+            just('r').to('\r'),
+            just('t').to('\t'),
+            just('u').ignore_then(text::digits(16).exactly(4).to_slice().validate(
+                |digits, e, emitter| {
+                    char::from_u32(
+                        u32::from_str_radix(digits, 16)
+                            .expect("Only valid digits should have been parsed in text::digits"),
+                    )
```
Gri-ffin, line 34: Wouldn't it be better to use chumsky here too? wouldn't this be hell to debug as a user? or is there a reason to crash here?

Gri-ffin, line 34: Wouldn't it be better to use chumsky here too and exit gracefully? wouldn't this be hell to debug as a user? or is there a reason to crash here?

mierak: The thing is, this invariant should be ensured by this `text::digits(16)` a few lines above. 

This panic should never really happen and if it does it should get fixed asap. But I suppose a more graceful error is fine too here.

mierak:
> @mierak I've been reviewing the PR and the discussions, and this looks really great. That said, I want to challenge a few aspects to explore potential pitfalls, if you don’t mind:

First of all, thank you for taking your time to look at this!

> **1. Robust Error Handling and User Feedback**
> 
> With this new "mini-language," I think strong error reporting becomes crucial to maintaining a good user experience. While technical users might be able to debug most issues, less technical users, especially once config files become more complex, may struggle to understand vague or cryptic errors.
> 
> This might be less of an issue now since the majority of users are on Linux and probably more technical, but I think you should still consider when users who are not much technical might use rmpc when the project become very popular.

Yup! Definitely a good point, chumsky has support for error reporting so I need to look into that.

> **2. Feature Scope and Complexity**
> 
> Features like `%padl`, `%if`, and `%progress` are exciting and open the door to powerful customizations. However, I suggest a slow approach here. You should the current core templating syntax first, and then gradually introduce more advanced functions after you gather user feedback.
> 
> I believe adding too many features too early might lead to a higher maintenance cost in the long run.

Well these are more or less already coming in, the `Truncate` transform is in and would be `%trunc` or `%truncate` with the current plan. This mini-language really only translates to the existing properties and has no logic of its own. But I do agree that a special care needs to be taken around these.

> **3. Documentation**
> 
> In my eyes, this is the most critical piece moving forward. Since you’re introducing a whole new syntax, a clear and complete documentation is very much essential for adoption. It’s very very important that users can quickly understand:
> 
>     * What syntax is valid and what isn’t
> 
>     * What properties exist and in what contexts they’re allowed
> 
>     * How groups, default fallbacks, and styles work
> 
>     * Error examples and how to fix them
> 
>     * Common patterns and idioms (e.g., fallback chaining, layout composition)
> 
> 
> This will make it much easier for users to adopt and use the language and even contribute safely.

I have been meaning to rework docs on all this for quite a while, the properties are a mess in general with them being documented in the `Header` section and this PR is probably the golden opportunity to finally do it.

src/config/theme/parser.rs
```diff
@@ -0,0 +1,891 @@
...
+                    just('b').to('\x08'),
+                    just('f').to('\x0C'),
+                    just('n').to('\n'),
+                    just('r').to('\r'),
+                    just('t').to('\t'),
+                    just('u').ignore_then(text::digits(16).exactly(4).to_slice().validate(
+                        |digits, e, emitter| {
+                            char::from_u32(u32::from_str_radix(digits, 16).expect(
+                                "Only valid digits should have been parsed in text::digits",
+                            ))
```
Gri-ffin, line 61: Same thing, here i think its best to use chumsky for error reporting and exit gracefully

src/config/theme/queue_table.rs
```diff
@@ -160,7 +168,20 @@ impl TryFrom<QueueTableColumnsFile> for QueueTableColumns {
...
+                            .into_result()
+                            .map_err(|e| {
+                                anyhow::anyhow!("Failed to parse property format: {:?}", e)
+                            })?
+                            .into_iter()
+                            .next()
+                            .context("A song property must be specified")?
+                            .try_into()?,
+                        (None, Some(prop)) => prop.try_into()?,
+                        (None, None) => panic!(""),
```
Gri-ffin, line 183: this would be a nightmare to debug as a user no? bailing would be better or atleast a panic with a message?

mierak: Thats just a left over from quickly getting this to work, take the PR with a grain of salt, its by no means close to finished.

roxwize:
Hi, I noticed how much activity this is getting. Sorry for not meaningfully pitching in since this was created, I'll be coming home from vacation in a couple days and by then I'll come around and actually fully look into this and offer my thoughts. So far this looks very promising.

I assume it's too late for breaking syntax changes but I will say that I think improvements could still be made for readability. Not to imply that all aspects of the language should be made extremely streamlined immediately, I just feel that at the moment there're a lot of tiny UX things here that still feel somewhat ambiguous/confusing. For one, all the dollar symbols are hard to follow, foobar's templating engine used various symbols to differentiate between different functionalities and it seems to be as terse as you can get with this sort of thing and I wonder if some notes could still be taken from that. I've brought that up multiple times before, though, so I won't repeat myself too much here.

At any rate, good luck with the implementation and thank you for going through with this. I'll have a look at the code when I get home.

mierak:
> 1. Robust Error Handling and User Feedback

Experimenting with chumsky's companion crate [ariadne](https://github.com/zesterer/ariadne). We can get pretty nice error reports on the formats (dont mind the missing component message). It would be nice if I could get list of supported properties out of it, but this is already pretty good imho.
image

Though it will have some issues on very small terminal sizes, but I think its still worth having.

mierak:
I introduced some shorter variants for the properties, for example "artist" and "ar" are equivalent. I have also decided to namespace the properties to avoid current and future conflicts. Still undecided on the namespacing, opinions anyone?

For string, $ is no longer needed, simply surrounding text in double or single quotes is fine, same for groups, no dollar sign needed, only brackets.

Introduced the truncate transform as either `%trunc` or `%truncate`.

Attributes on properties are now `snake_case` to be inline with the rest of the config.

And some general cleanup. Whats left is mostly:

* Docs - thats probably bulk of the work
* Tests - some are there but not nearly enough
* Finalize config deserialization and conversion
* Decide what to do with style modifiers, I find `mods: underlined | bold` kinda ugly.
* Cleanups and refactors of the code

example of the current syntax:
`format: "$st:elapsed' / '$st:duration{mods: bold}[' ('$st:bitrate' kbps)']",`
and the equivalent shortened form
`format: "$st:el' / '$st:dur{mods: bold}[' ('$st:br' kbps)']",`

mierak:
So this is basically done but hit a bit of a roadblock https://github.com/ron-rs/ron/issues/577

The current implementation with this issue will break existing configs if they use stuff like `Property(Status(StateV2()))` but `Property(Status(StateV2( )))`(not the space between parens) is fine.

Either we introduce a new set of options (browser_song_format_v2, format instead of content for Property panes, etc) like it was in the first place, or we wait whether this will get fixed at some point. I dont see any other options.

v-Nyo:
What would the disadvantage be if we introduce a new set of options? I am not sure I understand what you mean with this

>  format instead of content for Property panes, etc) like it was in the first place

mierak:
> What would the disadvantage be if we introduce a new set of options? I am not sure I understand what you mean with this

If the issue I reported didnt exist, then we could have `browser_song_format` accept both 
```ron
browser_song_format: "$file"
```
and 

```ron
browser_song_format: [(kind: Property(Filename))]
```
But without it getting fixed we would need `browser_song_format` and `browser_song_format_v2` for one or the other or accept that existing config files will get broken.

Same goes for the other places: song table format, property panes, tabs and layout.

And the disadvantage to introducing new options: Just that, its a new set of options that does not need to exist, its an additional complexity that *can* be avoided.

v-Nyo:
I see. 

I think temporally adding new options appended with something like "_templating" (easy find and replace) would be good until those issues you linked are fixed.  If you think this is worth it.

mierak:
> temporally adding

I personally would not mind, but once something is added and used by anyone it will then be a breaking change to their config to remove the option again and I want to avoid those as much as possible.

Issue #312, opened 2025-05-01
AddRandomItems


ptwales:
### Description

`ncmpcpp` has a feature called `AddRandomItems` by default bound to `` ` ``.  It prompts for songs, artists, albumArtists, or albums, then a number, and then adds a random number of those to the playlist from the entire library.  `rmpc` doesn't have anything like this.

[ncmpcpp's implementation](https://github.com/ncmpcpp/ncmpcpp/blob/e8f1c4bd48350789300c5f87c5ea21ea4256cf6f/src/actions.cpp#L2218)

### Proposed solution

Adding a command `addrandom <song|artist|albumArtist|album> <number>` command would be a simpler start than figuring out a full prompting dialog.  Users could then key-bind specific calls to it or execute it like `:addrandom song 50`.

mierak:
I will look into this, sounds fun.

mierak:
Some first part of the implementation here https://github.com/mierak/rmpc/pull/316 if you want to check it out, still undecided on a dialog for this but I think I will try to create something.

should work exactly as you described, for example:
```bash
rmpc addrandom album 3
```

mierak:
I have added a very simple modal that lets you add random songs from rmpc itself (`AddRandom` global action by default bound to `R`).

Feel free to leave a comment here if you have any feedback and thanks again.

PR #316, opened 2025-05-04
feat: addrandom command

mierak:
fixes https://github.com/mierak/rmpc/issues/312

### DIFF
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 
 use clap::{Parser, Subcommand, ValueEnum, ValueHint};
+use strum::IntoStaticStr;
 
 #[derive(Parser, Debug)]
 pub struct Args {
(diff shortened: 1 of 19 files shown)


src/config/cli.rs
```diff
@@ -19,9 +19,23 @@ pub struct Args {
...
+    AlbumArtist { count: usize },
+    Genre { count: usize },
+}
+
 #[derive(Subcommand, Clone, Debug, PartialEq)]
 #[clap(rename_all = "lower")]
 pub enum Command {
+    AddRandom {
+        #[command(subcommand)]
+        command: AddRandom,
```
ptwales, line 37: Using a subcommand here seems odd.  It results in a help output like this

```
Usage: rmpc addrandom <COMMAND>

Commands:
  song
  artist
  album
  albumartist
  genre
  help         Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

and then you have to drill into each one to figure out the entire command.  To me, it only makes sense if you want the possibility of having different arguments for each command.  

If you declared `AddRandom { tag: AddRandom, count: usize }`, you get all the help in one output like this, which I think is more helpful.

```
Usage: rmpc addrandom <TAG> <COUNT>

Arguments:
  <TAG>    [possible values: song, album, album-artist, artist, genre]
  <COUNT>

Options:
  -h, --help  Print help
```

mierak: You are completely right, I was trying to force the count positional arg after the subcommand but it did not occur to me that not using subcommand at all is the solution. Thanks.

PR #317, opened 2025-05-09
feat(lyrics): Implement word wrapping for lyrics pane.

xTayEx:
When displaying long lyrics line, current implementation will truncate the line if it‘s longer than the pane width. (as shown below).
image

With this PR, the long lyrics line will be wrapped.
image

### DIFF
--- a/src/ui/panes/lyrics.rs
+++ b/src/ui/panes/lyrics.rs
@@ -43,14 +43,21 @@ impl Pane for LyricsPane {
         let areas = Layout::vertical((0..rows).map(|_| Constraint::Length(1))).split(area);
         let middle_row = rows / 2;
 
-        let mut i: usize = 0;
-        while i < (rows as usize) {
-            let Some(idx) = (current_line_idx + i).checked_sub(middle_row as usize) else {
-                i += 1;
+        let mut area_idx: usize = 0;
+        let mut lyrics_fetch_idx: usize = 0;
+        while area_idx < (rows as usize) {
+            if (current_line_idx + area_idx).checked_sub(middle_row as usize).is_none() {
+                area_idx += 1;
+                continue;
+            }
+
+            let Some(idx) = (current_line_idx + lyrics_fetch_idx).checked_sub(middle_row as usize) else {
+                lyrics_fetch_idx += 1;
                 continue;
             };
+
             let Some(line) = lrc.lines.get(idx) else {
-                i += 1;
+                lyrics_fetch_idx += 1;
                 continue;
             };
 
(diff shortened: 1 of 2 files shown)


src/ui/panes/lyrics.rs
```diff
@@ -43,24 +43,36 @@ impl Pane for LyricsPane {
...
+                i += 1;
                 continue;
             };
             let Some(line) = lrc.lines.get(idx) else {
+                i += 1;
                 continue;
             };
 
-            let darken = (middle_row as usize).abs_diff(i) > 0 || !first_line_reached;
+            let wrapped_line: Vec<Line> = textwrap::wrap(&line.content, area.width as usize)
```
mierak, line 57: you can iterate directly over `textwrap::wrap(&line.content, area.width as usize)` below in the for loop, there is no need to wrap this in a Line and allocate a vec.

mierak:
This seems to skip over wrapped lines for me. See:
image
and after making the window smaller:
image
the first line is correctly wrapped, the second line gets skipped

xTayEx:
@mierak Thank you for your review! The latest code should fix the above problem (see below). I have refactored the lyrics rendering logic. Current method will first render the lyrics on the middle row, then expand bidirectionally. This can help to make it more easy to handle word wrapping, and reduce some complex index calculation.

Looking forward to you further feedback!

image

xTayEx:
Thank you for the feedback! I would like to ensure I fully understand the issue you mentioned.

In your example, when the current lyric is wrapped into multiple lines but cannot fully fit below the middle of the pane, the space above the middle line should be utilized to display more lines of the current lyric. This suggests prioritizing **showing as much of the current lyrics as possible**, even if it means the current lyric's first line may not be located at the middle.

My current implementation:
* The first line of the wrapped current lyric is placed at the `height / 2` (middle line), with subsequent lines rendered downward.
* If there's insufficient space below, these lines may be truncated (as seen in your example).

Your suggestion implies:
* If space above the middle line is available, it should be used to render the current lyrics instead of the previous ones, even if this shifts the lyrics block upward (away from strict middle alignment).

So the key question is:
Should the current lyric's lines be allowed to shift upward to maximize visibility, even if this means breaking the strict middle alignment?

mierak:
> Should the current lyrics' lines be allowed to shift upward to maximize visibility, even if this means breaking the strict middle alignment?

That is what I meant yeah. The way I see it the most important line is the current lyrics one and should have the most focus.

But again, this might be opening a new whole can of worms so I am fine with leaving it as it is now.

xTayEx:
Got it - I agree that this is a new topic, but we can leave it as it is now. Happy to explore it further when needed. Also, changelog has been updated for this PR.

mierak:
merging this then, thank you

PR #329, opened 2025-05-23
feat: Add Insert and InsertAll commands

tarneaux:
### Description
The added Insert and InsertAll commands allow adding songs directly after the current song, with mpd's POSITION argument on its API.

### Scope
This just changes what it needs to, it was done by adding the command and then being taken by the hand by the compiler and adding the explicit matches everywhere they were needed. Basic testing was done.

### Criticism
Since I don't know exactly how the queue view works, I didn't implement its Insert action, even though it would be interesting, since it is often needed to add a song already in the queue directly after the current track. Note it is still possible to do so manually by simply moving the track up. An implementation of this would probably add the song a second time instead of moving it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ lists they are listed as multiple entries.
 - introduced `preview_label_style` and `preview_metadata_group_style` in theme config
 - added support for soundcloud to `addyt`
 - added `AddReplace` and `AddAllReplace` actions which work smimilarly to `Add` and `AddAll` but replace the current queue instead of appending
+- added `Insert` and `InsertAll` actions which work similarly to `Add` and `AddAll` but insert after the playing song
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ strum = { workspace = true }
 clap_complete = "4.5.45"
 clap_mangen = "0.2.26"
 vergen-gitcl = { version = "1.0.5", features = ["build"] }
+anyhow = "1.0.95"
 
 [dev-dependencies]
 rstest = "0.24.0"
--- a/build.rs
+++ b/build.rs
@@ -10,6 +10,15 @@ use clap_complete::{
 use clap_mangen::Man;
 use vergen_gitcl::{Emitter, GitclBuilder};
 
+// Mock mpd::QueuePosition for cli.rs to be able to import it. Also see header
+// comment in cli.rs.
+#[path = "src/mpd/queue_position.rs"]
+mod queue_position;
+
+mod mpd {
+    pub use super::queue_position::QueuePosition;
+}
+
 static NAME: &str = "rmpc";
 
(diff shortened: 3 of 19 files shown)


tarneaux:
Note that I made no changes to the documentation since I'm unsure which files I should edit. Basically the only thing to add is a small description of the Insert and InsertAll commands in the keybindings page.
Maybe the default configuration should also include keys like <C-a> and <C-A> for these two actions.

mierak (commented):
> Since I don't know exactly how the queue view works, I didn't implement its Insert action, even though it would be interesting, since it is often needed to add a song already in the queue directly after the current track. Note it is still possible to do so manually by simply moving the track up. An implementation of this would probably add the song a second time instead of moving it.

I dont think the queue implementation is needed, though if you want to, here are some pointers:

The current queue is stored in the context because it needs to be accessed in places other than the Queue pane.
For the insert function you would do something like this:
```rust
CommonAction::Insert => {
    let song_under_cursor =
        self.scrolling_state.get_selected().and_then(|idx| context.queue.get(idx));

    if let Some(song) = song_under_cursor {
        let file = song.file.clone();
        context.command(move |client| {
            client.add(&file, true)?;
            Ok(())
        });
    }
}
```
You probably need to only handle Insert and Insert with marked items, I dont see how insert all makes sense in the queue.

Other than the few comments I left it looks good, thank you!

mierak:
> Note that I made no changes to the documentation since I'm unsure which files I should edit. Basically the only thing to add is a small description of the Insert and InsertAll commands in the keybindings page.

I think just the [keybinds](https://mierak.github.io/rmpc/next/configuration/keybinds/) is fine. There is an `Edit page` button at the bottom of each page if you have trouble finding the corresponding mdx file. But generally you only ever want to edit files in `docs/src/content/docs/next` as that is the section of the docs for the git version.

> Maybe the default configuration should also include keys like and for these two actions.

No strong opinion on this, feel free to set some defaults if you want to, but the default keys are pretty crowded by now.

tarneaux:
Feel free to squash my commits into the first one.

mierak:
How do you supply `-1` for example to `rmpc add`? Clap has some issues inferring values starting with hyphen as argument. There are some workarounds like https://docs.rs/clap/latest/clap/struct.Arg.html#method.allow_negative_numbers but I think having the insert boolean in CLI was fine, it was mostly the client I was concerned about.

Other than this it looks ok to me. `Insert` fails when no song is currently playing/paused, but the error message is descriptive enough so thats fine to leave as is unless you want to spend more time on it.

tarneaux:
Turns out I had completely messed up my string indexes. I've fixed it now and added allow_negative_numbers for these args. Theoretically I could also add the insert boolean again, but I feel like having the full control is better, since the format is quite easy to understand (I mean, it's not like you have to write an entire JSON object just to insert).

I may make a fix for the Insert in an empty queue, but for now I'll leave the PR as is.

mierak:
great, thanks!

PR #349, opened 2025-05-25
perf: skip calculating rows that are not visible in queue

mierak:


soifou:
No more "weirdness" on my side either, although I haven’t done a thorough check. I’ll let you know if I notice any other regressions. Thanks again for the prompt fix!

PR #373, opened 2025-05-28
Feat: sort keybinds in the help modal

osamaaj03:
Idk at least the items don't get shuffled randomly with each initialization. Also, I noticed a typo in the chain.


mierak:
Yep, its better than not having it, thanks. Lets keep that issue open though.

PR #377, opened 2025-05-28
feat: add block album_art support

Sang-it:
Add block album_art.
Copied over ideas from [viu](https://github.com/atanunq/viu) to add block print support. Should work everywhere.
Tested on Alacritty.

It was annoying that Alacritty didn't have image support. This atleast gives it some image rendering to default to.

Screenshot 2025-05-28 at 5 37 31 PM

### DIFF
--- a/src/ui/image/block.rs
+++ b/src/ui/image/block.rs
@@ -1,0 +1,198 @@
+use super::{AlbumArtConfig, Backend, ImageBackendRequest, clear_area};
+use crate::shared::ansi_colors::ansi256_from_rgb;
+use crate::{
+    config::{
+        Size,
+        album_art::{HorizontalAlign, VerticalAlign},
+    },
+    shared::{
+        image::{AlignedArea, resize_image},
+        macros::{status_error, try_cont},
+    },
+    try_skip,
+    ui::image::{EncodeRequest, facade::IS_SHOWING, recv_data},
+};
+use anyhow::{Result, bail};
+use crossbeam::channel::{Sender, unbounded};
+use crossterm::style::{ResetColor, SetBackgroundColor, SetForegroundColor};
+use crossterm::{
+    cursor::{MoveTo, RestorePosition, SavePosition},
+    queue,
+    style::{Color, Colors, Print},
+};
+use image::{DynamicImage, GenericImageView, Rgba};
+use ratatui::layout::Rect;
+use std::io::Write;
+use std::sync::{Arc, atomic::Ordering};
+
+#[derive(Debug)]
+pub struct Block {
+    sender: Sender<ImageBackendRequest>,
+    colors: Colors,
+    handle: std::thread::JoinHandle<()>,
+}
+
+impl Backend for Block {
+    fn hide(&mut self, size: Rect) -> anyhow::Result<()> {
+        clear_area(&mut std::io::stdout().lock(), self.colors, size)
+    }
+
+    fn show(&mut self, data: Arc<Vec<u8>>, area: Rect) -> Result<()> {
+        Ok(self.sender.send(ImageBackendRequest::Encode(EncodeRequest { area, data }))?)
+    }
+
+    fn set_config(&self, config: AlbumArtConfig) -> Result<()> {
+        Ok(self.sender.send(ImageBackendRequest::SetConfig(config))?)
+    }
+
+    fn cleanup(self: Box<Self>, _area: Rect) -> Result<()> {
+        self.sender.send(ImageBackendRequest::Stop)?;
+        self.handle.join().expect("block thread to end gracefully");
+        Ok(())
+    }
+}
+
+const LOSS_FACTOR: usize = 2;
+const MAX_WIDTH: u16 = 180;
+const PRINT_BLOCK: &str = "\u{2580}";
+
+#[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+impl Block {
+    pub(super) fn new(config: AlbumArtConfig) -> Self {
+        let (sender, receiver) = unbounded::<ImageBackendRequest>();
+        let colors = config.colors;
+
+        let handle = std::thread::Builder::new()
+            .name("block".to_string())
+            .spawn(move || {
+                let mut config = config;
+                let mut pending_req = None;
+                'outer: loop {
+                    let EncodeRequest { data, area } =
+                        match recv_data(&mut pending_req, &mut config, &receiver) {
+                            Ok(Some(msg)) => msg,
+                            Ok(None) => break,
+                            Err(err) => {
+                                log::error!("Error receiving ImageBackendRequest message: {err}");
+                                break;
+                            }
+                        };
+
+                    // consume all pending messages, skipping older encode requests
+                    for msg in receiver.try_iter() {
+                        match msg {
+                            ImageBackendRequest::Stop => break 'outer,
+                            ImageBackendRequest::SetConfig(cfg) => config = cfg,
+                            ImageBackendRequest::Encode(req) => {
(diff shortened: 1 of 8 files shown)


mierak (commented):
Ok, this is sick. Few more things aside from the comments though:

* Please make a note in the changelog when you are done
* [docs](https://mierak.github.io/rmpc/next/configuration/album-art/) are needed
* Why is the `ansi_colours` vendored in? I see it available on crates.io
* From what I read, its fine to include LGPL code assuming only the code in question is LGPL, right?
* can we make it a fallback after ueberzug before returning None in `shared::image::determine_image_support`?
* There are also clearing/rerendering issues when opening and closing a modal window, for example the `ShowHelp`, bound to `~` by default.

The album arts are very finicky unfortunately because we are completely bypassing ratatui's rendering pipeline here, that is why the stdout locks are in place.

Sang-it:
Let me see if can fix these errors.
That was quick feedback.

src/ui/image/block.rs
```diff
@@ -122,6 +117,110 @@ impl Block {
...
+        VerticalAlign::Bottom => available_area.bottom().saturating_sub(new_height),
+    };
+
+    let result = AlignedArea {
+        area: Rect::new(new_x, new_y, new_width, new_height),
+        size_px: Size {
+            width: (new_width * 2).min(max_size_px.width),
+            height: (new_height * 2).min(max_size_px.height),
+        },
+    };
```
Sang-it, line 187: I had to do this to make the things line up properly. Let me know if you find more bugs. I fiddled around with it, everything seems to work fine.

src/ui/image/block.rs
```diff
@@ -0,0 +1,288 @@
...
+    resized_area: AlignedArea,
+) -> std::io::Result<()> {
+    queue!(w, SavePosition)?;
+    queue!(w, MoveTo(resized_area.area.x, resized_area.area.y))?;
+
+    let (i_height, i_width) = img.dimensions();
+    let img_buffer = img.to_rgba8();
+
+    let mut y_step = 1;
+    for y in (0..i_height).step_by(2) {
```
mierak, line 249: Looking at this function closer, you are iterating over pixels in the image. Instead you should iterate over cells (width and height from resized_area). It should also be more performant because way fewer iterations are needed.

The image size in pixels and `resized_area` size should be the same - meaning `resized_area.width * terminal_cell_width == img.width` Then in the loop over columns/rows you choose color from pixels which lie within the cell (either choose randomly or average them, either is fine).

Suggested changes:

* Delete the copied `resize_image` and use existing one from shared
* Delete the copied `create_aligned_area` and use existing one from shared
* Iterate over cells instead of pixels
* For each cell choose a color from corresponding pixels in the image - either average, take one randomly, quantize them, etc...

After aplying these changes (Please take this code with a grain of salt, the casts need to be redone etc, but it works):
```rust
    let (cell_width, cell_height) = (
        resized_area.size_px.width as f64 / resized_area.area.width as f64,
        resized_area.size_px.height as f64 / resized_area.area.height as f64,
    );
    for y in 0..resized_area.area.height {
        for x in 0..resized_area.area.width {
            let top = match img_buffer
                .get_pixel_checked(x as u32 * cell_width as u32, y as u32 * cell_height as u32)
            {
                Some(rgb) => rgb,
                None => &image::Rgba([0, 0, 0, 255]),
            };
            let bottom = match img_buffer.get_pixel_checked(
                x as u32 * cell_width as u32,
                y as u32 * cell_height as u32 + (cell_height / 2.0) as u32,
            ) {
                Some(rgb) => rgb,
                None => &image::Rgba([0, 0, 0, 255]),
            };
            let fg = color_from_pixel(*top);
            let bg = color_from_pixel(*bottom);
            queue!(w, SetForegroundColor(fg), SetBackgroundColor(bg), Print(PRINT_BLOCK))?;
        }
        queue!(w, MoveTo(resized_area.area.x, resized_area.area.y + y))?;
    }
```
This also fixes the aspect ratio issue I mentioned and makes the alignment work the same as the rest of the image backend.

See example:
image

mierak (commented):
1. I am still getting artifacts when opening and closing a modal window(ShowHelp) over the album art

<details>

<summary>expand: before and after</summary>

image
image
</details>

2. The aspect ratio is wrong - the image is squished. Its also not centered properly (which is probably just a symptom of the aspect ratio being wrong). I provided the reason why and how to fix it in the comments.
<details>

<summary>Block</summary>

image
</details>
<details>

<summary>Kitty</summary>

image
</details>

src/ui/image/block.rs
mierak: This should not be needed and is wrong even, because of reasons I outlined below. This whole function should be removed.

Sang-it:
I rewrote the block display method to accurately match viu.

- Naive
Screenshot 2025-05-29 at 10 49 26 PM


- Viu Method
Screenshot 2025-05-29 at 10 49 33 PM

- And the modal issue is fixed too. I had to always fetch the image for some reason. I have no idea why. Maybe some weird race condition. But it works. It shouldn't be an issue as Image fetches are like ~2ms.

src/ui/panes/album_art.rs
```diff
@@ -140,7 +140,7 @@ impl Pane for AlbumArtPane {
             UiEvent::ModalClosed if is_visible => {
                 self.is_modal_open = false;
 
-                if self.fetch_needed {
+                if self.fetch_needed || context.config.album_art.method == ImageMethod::Block {
```
mierak, line 143: This is not the correct solution, the same problem exists in other places and this fixes only the modal. The modal closing is just easiest to reproduce.
```suggestion
                if self.fetch_needed {
```

src/ui/image/block.rs
```diff
@@ -0,0 +1,203 @@
...
+                        clear_area(&mut w, config.colors, area),
+                        "Failed to clear image area"
+                    );
+
+                    let (img, resized_area) = try_cont!(
+                        resize_image(&data, area, config.max_size, config.halign, config.valign),
+                        "Failed to resize block image"
+                    );
+
+                    try_skip!(display(&mut w, &img, resized_area), "Failed to display block");
```
mierak, line 106: Your solution made me realize what is different between the block impl and the others. You are resizing the image AFTER you lock the stdout and check IS_SHOWING, the resizing is a long operation. This should reordering of things fixes it.

```suggestion
                    let (img, resized_area) = try_cont!(
                        resize_image(&data, area, config.max_size, config.halign, config.valign),
                        "Failed to resize block image"
                    );

                    // consume all pending messages, skipping older encode requests
                    for msg in receiver.try_iter() {
                        match msg {
                            ImageBackendRequest::Stop => break 'outer,
                            ImageBackendRequest::SetConfig(cfg) => config = cfg,
                            ImageBackendRequest::Encode(req) => {
                                pending_req = Some(req);
                                log::debug!(
                                    "Skipping image because another one is waiting in the queue"
                                );
                                continue 'outer;
                            }
                        }
                    }

                    let mut w = std::io::stdout().lock();
                    if !IS_SHOWING.load(Ordering::Relaxed) {
                        log::trace!(
                            "Not showing image because its not supposed to be displayed anymore"
                        );
                        continue;
                    }

                    try_cont!(
                        clear_area(&mut w, config.colors, area),
                        "Failed to clear image area"
                    );

                    try_skip!(display(&mut w, &img, resized_area), "Failed to display block");
```

mierak (commented):
> And the modal issue is fixed too. I had to always fetch the image for some reason. I have no idea why. Maybe some weird race condition. But it works. It shouldn't be an issue as Image fetches are like ~2ms.

Yeah, its a race condition, the changes I provided should solve it though.

PR #392, opened 2025-05-31
Fix: Lyrics files with ']' in the metadata fields not getting indexed and title not getting matched

Lxtharia:
I've noticed that titles that include a closing bracket (like "some song title [Explicit]") didn't get indexed.
During debugging this, I found that the artist metadata field is passed twice to the find_lrc function.

This PR fixes this.

I had some more issues trying to get my .lrc files to be read/matched, though they might be too specific/uncommon to bother fixing:
- Files that start with [BOMs](https://de.wikipedia.org/wiki/Byte_Order_Mark) fail to be parsed
- Length mismatches, which made me wonder if matching the length is truly nessesary
- Some .lrc files missing an album metadata field, resulting in them not getting indexed

### DIFF
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -70,7 +70,7 @@ impl LrcIndex {
     pub fn find_lrc_for_song(&self, song: &Song) -> Result<Option<Lrc>> {
         match (
             song.metadata.get("artist"),
-            song.metadata.get("artist"),
+            song.metadata.get("title"),
             song.metadata.get("album"),
             song.duration,
         ) {
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -134,11 +134,8 @@ impl LrcIndexEntry {
             let metadata = buf
                 .trim()
                 .strip_prefix('[')
+                .and_then(|s| s.strip_suffix(']'))
                 .with_context(|| format!("Invalid lrc line format: '{buf}'"))?;
-            if !metadata.ends_with(']') {
-                break;
-            }
-            let metadata = &metadata[..metadata.len() - 1];
 
             match metadata.chars().next() {
                 Some(c) if c.is_numeric() => {
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -81,7 +81,7 @@ impl LrcIndex {
                 match lrc_opt {
                     None => log::trace!("No Lyrics found for {:?}", song.metadata),
                     Some(lrc) => log::trace!("Lyrics found at {:?}", lrc.path),
-                };
+                }
                 lrc_opt
             }
             _ => None,
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -77,7 +77,12 @@ impl LrcIndex {
             (Some(artist), Some(title), Some(album), length) => {
                 // TODO xxx.last() is called here to not change existing behavior. Consider
                 // supporting all the tag entries
-                self.find_lrc(artist.last(), title.last(), album.last(), length)
+                let lrc_opt = self.find_lrc(artist.last(), title.last(), album.last(), length);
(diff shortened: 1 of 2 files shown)


src/shared/lrc/index.rs
```diff
@@ -123,14 +131,14 @@ impl LrcIndexEntry {
...
+            let metadata = buf
                 .trim()
                 .strip_prefix('[')
-                .and_then(|s| s.split_once(']'))
                 .with_context(|| format!("Invalid lrc line format: '{buf}'"))?;
-            if !rest.is_empty() {
+            if !metadata.ends_with(']') {
                 break;
             }
+            let metadata = &metadata[..metadata.len() - 1];
```
mierak, line 141: wouldn't changing split_once to strip_suffix and doing break on None be enough instead of indexing into the string?

mierak:
> Files that start with [BOMs](https://de.wikipedia.org/wiki/Byte_Order_Mark) fail to be parsed

I mostly went by what was written on wiki, there does not seem to be a formal specification for lrc files and I didnt even know that BOM is legal in lrc.

> Length mismatches, which made me wonder if matching the length is truly nessesary

There is 3 seconds leeway in the length matching. I am fine with increasing this, but to me it still makes sense. Say you have extended version of a song but lyrics only for the shorter version.

> Some .lrc files missing an album metadata field, resulting in them not getting indexed

I was thinking about removing this one as well, or at least making it more lax

src/shared/lrc/index.rs
Lxtharia: Ah, wasn't aware of that function!

mierak:
```rust
            let (metadata, rest) = buf
                .trim()
                .strip_prefix('[')
                .and_then(|s| s.rsplit_once(']'))
                .with_context(|| format!("Invalid lrc line format: '{buf}'"))?;
            if !rest.is_empty() {
                break;
            }

```
this should work, simply splitting at last occurence of ] instead of the first

Lxtharia:
> and I didnt even know that BOM is legal in lrc.
Maybe it isn't! I've never heard of BOM before today tbh. and I've just collected a few .lrc files from when I used Minilyrics on windows, so one of those two might be at fault.

I'll probably just have to `sed` over all of them once and solved that issue, but still, never had any issues with this before.

> Say you have extended version of a song but lyrics only for the shorter version.
Yeah, I get that! But on the other hand, there might be versions with extra silence at the end or end parts getting cut, where the lyrics stay the same. It would be best to only consider length as a criteria when there is more than one matches.

The same can be maybe applied to case sensitivity and albums.
Some .lrc creators might write `A-ha` instead of `a-ha` as the artists name, and some songs might appear in different albums while being the same song.
It's not easy to judge when a not-exact-match is still correct and expected, but in my opinion I think it's more frustrating to not have lyrics displayed at all than having the extended-version-lyrics on my normal-version song playing.

mierak:
> It would be best to only consider length as a criteria when there is more than one matches.

> The same can be maybe applied to case sensitivity and albums.

both of these make sense to me!

> It's not easy to judge when a not-exact-match is still correct and expected

yeah, this whole matching and parsing implementation was all guesswork on my side because there does not seem to be one stadardized way :disappointed: 

Anyway, this seems good to go to me.

Lxtharia:
> yeah, this whole matching and parsing implementation was all guesswork on my side because there does not seem to be one stadardized way 😞

understandable!
In my experience it seems that one of the big issues is also just data mismatch. Different streaming platforms having slightly different titles or multiple releases of the same song with "(feat. Artist)" as part of the title in one version and as part of the artists in the other version, as well as 3 different lyrics providers that all have their own conventions, etc...

And in this context of matching a song not only with all locally available lyrics files but with a enormous database of all known lyrics files, the existing matching algorithms are probably way more advanced but still not perfect.

I kinda forgot where I wanted to go with this.
"No code can fix messy data" probably

PR #393, opened 2025-06-01
Add lyrics style to theme

khalil-cheddadi:
hello, i tried to add lyrics styles to the theme folder, for now only two properties which are active_line_style and line_style
it seems everything is working properly
although i couldn't figure out how to get `rmpc theme` to add the new config part to the output
also i couldn't find a way to add default modifers

but now you can do something like this:

image

### DIFF
--- a/src/config/theme/lyrics.rs
+++ b/src/config/theme/lyrics.rs
@@ -2,7 +2,7 @@ use anyhow::Result;
 use ratatui::style::{Color, Style};
 use serde::{Deserialize, Serialize};
 
-use super::{Modifiers, StyleFile, style::ToConfigOr};
+use super::{StyleFile, style::ToConfigOr};
 
 #[derive(Debug, Default, Clone)]
 pub struct LyricsConfig {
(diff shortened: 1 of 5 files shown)


mierak:
Theme is embedded from docs directly into the binary https://github.com/mierak/rmpc/blob/6b0e4661618805c4d24260214cbe56721b0842bc/src/main.rs#L57C1-L61C10

So for that you need to change the example in the docs, also please add these new settings to the theme section of the docs and mention the change in the changelog.

Other than that it looks fine!

src/config/theme/mod.rs
```diff
@@ -101,6 +104,7 @@ pub struct UiConfigFile {
     pub(super) format_tag_separator: String,
     #[serde(default)]
     pub(super) level_styles: LevelStylesFile,
+    pub(super) lyrics: LyricsConfigFile,
```
mierak, line 107: #[serde(default)] is missing here, without it existing configs will break

src/config/theme/lyrics.rs
```diff
@@ -0,0 +1,43 @@
...
+                modifiers: Some(Modifiers::Dim),
+            }),
+        }
+    }
+}
+
+impl LyricsConfigFile {
+    pub(super) fn into_config(self) -> Result<LyricsConfig> {
+        Ok(LyricsConfig {
+            active_line_style: self.active_line_style.to_config_or(Some(Color::Blue), None)?,
```
mierak, line 39: Oh, and one more thing, it would be better to let it fallback to the previous values (text_color/highlight_style) instead of blue/white? That way it will keep changeing with your config if you dont change lyrics explicitly.

khalil-cheddadi:
hi, i am kind of new to rust so i wasn't able to resolve everything you pointed out

mierak:
Feel free to ask any questions, I can give you some pointers.

Issue #391, opened 2025-05-31
Can't figure out how to get the lyrics to show


BeatlesNo1:
Hi,

Great project. I have one '.lrc' file currently as I am trying to get the lyrics to show up. Here is my setup: I have my '.ogg' music files in my 'Music' directory which has various sub directories based on language and genre. Also in that same directory i have a 'Lyrics' directory where i will hold all of my .lrc files. I checked and tags are matching. Here is my .lrc file:

`[ar:Simon & Garfunkel]
[ti:The Sound Of Silence]
[00:03.76] Hello darkness, my old friend
[00:07.62] I've come to talk with you again
[00:11.26] Because a vision softly creeping
[00:16.48] Left its seeds while I was sleeping
[00:21.26] And the vision that was planted in my brain
[00:27.86] Still remains
[00:30.74] Within the sound of silence
[00:37.24] In restless dreams, I walked alone
[00:41.90] Narrow streets of cobblestone
[00:46.19] 'Neath the halo of a streetlamp
[00:50.94] I turned my collar to the cold and damp
[00:55.20] When my eyes were stabbed by the flash of a neon light
[01:02.33] That split the night
[01:05.38] And touched the sound of silence
[01:11.41] And in the naked light, I saw
[01:16.16] Ten thousand people, maybe more
[01:20.35] People talking without speaking
[01:24.87] People hearing without listening
[01:29.33] People writing songs that voices never shared
[01:37.76] No one dared
[01:41.43] Disturb the sound of silence
[01:45.92] "Fools", said I, "You do not know
[01:50.55] Silence like a cancer grows
[01:55.10] Hear my words that I might teach you
[01:59.30] Take my arms that I might reach you"
[02:03.92] But my words like silent raindrops fell
[02:11.22] And echoed in the wells of silence
[02:18.44]
[02:22.13] And the people bowed and prayed
[02:25.19] To the neon god they made
[02:29.28] And the sign flashed out its warning
[02:33.78] In the words that it was forming
[02:38.25] And the sign said, "The words of the prophets are written on the subway walls
[02:45.70] In tenement halls"
[02:49.38]
[02:54.74] And whispered in the sounds of silence
[03:01.30]
` 

rmpc version 0.8.0.
CachyOS with Hyprland. Terminal is Alacritty.

Here is 'lyricsindex' result:

`{
  "index": []
}` 

I did specified 'lyrics_dir' variable in rmpc. Not in MPD.
[...]

PR #396, opened 2025-06-01
feat: make lyrics matching more lenient

mierak:
## Description

This makes lrc matching more lenient. Matching is now case insensitive and album is only considered if both the song and the lrc file contain album metadata, but if multiple matches are found, the matches with album are prioritized.

Length is now only considered with multiple matches:
* First try to find lrc that is +- 5 seconds
* If not found check whether any lrc file without length was found
* as a last resort, return the closest match by length

There is a bunch of tests which should hopefully describe the behavior in more detail.

@Lxtharia I am looking for a feedback since you were interested in this. What do you think about this behavior?

Note: I am merging this to get it off my head, but please do leave a feedback if you have any.

### Related issues

https://github.com/mierak/rmpc/issues/391
https://github.com/mierak/rmpc/pull/392

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ lists they are listed as multiple entries.
 - Refactored yt-dlp to make it easier to add support for more hosts
 - `scrollbar` theme option now also accepts `None` as a valid value to hide all scrollbars in rmpc
 - `TogglePause` in both the keybind and CLI to issue play if the current state is stopped
+- Made lyrics index matching more lenient
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Lxtharia:
Looks good!
This motivated me to fix my BOM'ed files and other quirky files from my library and besides the mentioned natural data mismatch (it's absolute chaos lmao) matching works well where I expect it to work!
Thank you for seeing to this so quickly, especially considering that I've only just known rmpc since yesterday :)

mierak:
Thats great to hear, thanks!
And yeah, there has been a huge influx of new users. Rmpc went from 300 stars to almost 800 in a span of a single week. I am still coming to terms with that :smile:

PR #406, opened 2025-06-03
Added Clear Error Modal Feature

### LINKED ISSUE #404: Clear error modal on config/theme update
### Description

1. Tweak config/theme
2. Get any error modal
3. Fix error
4. Error modal doesn't disappear 

### Proposed solution

Clear previous error modal on config/theme update

Thunder-Blaze:
## Description
- Added a new event that will clear the Config Modal when the config is reloaded and found correct

### Related issues
- Fixes #404

### Checklist

- [x] All tests passed

### DIFF
--- a/src/ui/mod.rs
+++ b/src/ui/mod.rs
@@ -481,13 +481,23 @@ impl<'ui> Ui<'ui> {
     pub fn on_ui_app_event(&mut self, event: UiAppEvent, context: &mut AppContext) -> Result<()> {
         match event {
             UiAppEvent::Modal(modal) => {
-                self.modals.push(modal);
+                if let Some(id) = modal.get_id() {
+                    if let Some(existing) =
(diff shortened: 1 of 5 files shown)


src/ui/mod.rs
```diff
@@ -485,6 +485,13 @@ impl<'ui> Ui<'ui> {
                 self.on_event(UiEvent::ModalOpened, context)?;
                 context.render()?;
             }
+            UiAppEvent::PopConfigErrorModal => {
+                if let Some(last_modal) = self.modals.last() {
```
mierak, line 489: I would even go as far as to pop all the `config_error` modals. Because we successfully read a config file at this point, meaning it must be fixed.

mierak: Loop through is fine. The performance difference is negligible, we are expecting single digits of modals at most and its not on any hot path anyway. Thanks!

Issue #298, opened 2025-04-10
ncmpcpp like visualisation


illegitimate-egg:
### Description

ncmpcpp had a set of music visualizers that rmpc doesn't have an answer to. This could be useful for monitoring the left and right channel during audio configuration.

mierak:
I experimented with this in the past but I could only get so far with my limited understanding of spectrum analysis. MPD provides [PCM data](https://mpd.readthedocs.io/en/latest/plugins.html#fifo) over fifo which can be used to run the analysis on.

Second thing to solve is how to render the visualization itself. High framerate is required, ratatui is an immediate mode UI and rendering the whole UI would be very wasteful if nothing changed. We probably want it to run on a different rendering pipeline than the rest of the application (similar to album art now).

Definitely something I will explore in the future again or would be very happy with contributions.

mierak:
https://github.com/mierak/rmpc/pull/415 please try it out if you have time, it changed a lot of things in the core functionality as well

mierak:
Cava support has landed. Please test it out, it has been quite a change :smile:.

osamaaj03:
this is really cool! Sorry for not getting back to you sooner—I’ve been super busy the last few days and will probably be for a while. But I’d love to test this out as soon as I can! Honestly, I didn’t realize it would take so much effort when I pitched the idea. At least now you have the code ready for when you finish your own visualization!

mierak:
Yeah, it was a lot because rmpc has to manage both the cava process lifecycle and manage the different rendering pipelines and make them not interfere with each other (ratatui, album art and now cava). There is also bunch of stuff like handling tmux switching so cava does not run when rmpc is not visible etc.

 Some of the rendering code can definitely be reused. I also want to implement some more display/theming options for the visualiser on rmpc's side, but that can wait a bit and should be fairly simple to do with all that is done now.

PR #415, opened 2025-06-05
feat: support cava visualiser

mierak:
## Description

This PR adds support for [cava](https://github.com/karlstav/cava) via a new `Cava` pane. Cava has to be installed on your system and on your `$PATH`.

A bunch of **testing is needed**, so please try this out if you are interested in the feature. Some things to look out for include opening and closing modals, tab switching, tmux integration, resizing of the terminal(resizing messes up everything, but should fix itself in a moment), interaction with all the different album art image backends, config reloading and different terminal emulators.

Some core functionality had to change, stdout is now behind a mutex because proper locking is needed with cava rendering so quickly and on a different pipeline from ratatui. parking_lot mutex is used instead of the std one because of its fairness guarantees.

This is not done yet, ~~configuration is not exposed at all~~, configuration reflects cava configuration itself, refer to their docs for more info, but not all options are supported, docs are missing, testing was done not well enough.

Example configuration, this in your mpd.conf
```conf
audio_output {
   type                    "fifo"
   name                    "my_fifo"
   path                    "/tmp/mpd.fifo"
   format                  "44100:16:2"
}
```

and this in config.ron
```ron
    cava: (
        framerate: 60,
        lower_cutoff_freq: 50,
        higher_cutoff_freq: 10000,
        input: (
            method: Fifo,
            source: "/tmp/mpd.fifo",
            sample_rate: 44100,
            channels: 2,
            samble_bits: 16,
        ),
        smoothing: (
            noise_reduction: 20,
            monstercat: true,
        ),
    ),
```

And finally, a preview:

https://github.com/user-attachments/assets/f60d7d7c-3502-4428-be92-3e15cbb4644d

### Related issues

#298 

### Checklist

- [ ] All tests passed


Thunder-Blaze:
Awesome !!!
Will have to get your config and theme (layout) :D
btw will test it out at night when I make the PR 🙂

mierak:
> Will have to get your config and theme (layout) :D

My configs are a complete mess because I change them all the time while working on rmpc. In the video I was using the debug config. You can find it here https://github.com/mierak/dotfiles/tree/chezmoi/dot_config/rmpc

>  btw will test it out at night when I make the PR 🙂

thanks!

mierak:
im having way too much fun with it (its configurable, as is tradition)

image

Thunder-Blaze:
great, It's working perfectly on mine too

image

Thunder-Blaze:
I changed the code to support gradient colors

| `/tmp/rmpc/cava.conf`   | `config/cava.rs`    |
| ---    | ---   |
| image | image |

but the gradient colors still not working, am I missing something ?
those colors are coming from `config.debug.ron`
Ig I'll wait for your colors commit

mierak:
> those colors are coming from config.debug.ron

yeah that is still very much work in progress which I have only locally, and the colors dont get passed to cava anyway, because the output rendering is completely rmpc's side and has nothing to do with cava which just provides the raw binary data.

Thunder-Blaze:
Oh, I see then there's no point in configuring colors for cava, need to configure colors in rmpc theme or smth since rmpc is rendering the visualizers based on data from cava

Thunder-Blaze:
Lol I should've read the whole changes first 🥲, found render_cava in panes which is responsible for coloring and theming

mierak:
> Lol I should've read the whole changes first 🥲, found render_cava in panes which is responsible for coloring and theming

Exactly. Essentially, we get a number between 0 and 65535 for each bar and are free to do whatever we want with it.

mierak:
this is basically done implementation wise, docs are still missing

mierak:
Tested in alacritty, wezterm, kitty, ghostty, foot. Everything seems to behave. Docs are in. This should be ready to go.

Thunder-Blaze:
Noice, time to make my autogen theme 😏

PR #416, opened 2025-06-05
Add lyrics config option to show line timestamps

khalil-cheddadi:
## Description

adds lyrics.show_timestamp to add the timestamp of each line
it ignores empty lines

image

### Checklist

- [x] All tests passed

### DIFF
--- a/src/config/theme/lyrics.rs
+++ b/src/config/theme/lyrics.rs
@@ -1,20 +1,17 @@
-use anyhow::Result;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Default, Clone)]
 pub struct LyricsConfig {
-    pub show_timestamp: bool,
+    pub timestamp: bool,
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
 pub struct LyricsConfigFile {
-    pub(super) show_timestamp: bool,
+    pub(super) timestamp: bool,
 }
 
-impl TryFrom<LyricsConfigFile> for LyricsConfig {
-    type Error = anyhow::Error;
-
-    fn try_from(value: LyricsConfigFile) -> Result<Self> {
-        Ok(LyricsConfig { show_timestamp: value.show_timestamp })
+impl From<LyricsConfigFile> for LyricsConfig {
+    fn from(value: LyricsConfigFile) -> Self {
+        LyricsConfig { timestamp: value.timestamp }
     }
 }
--- a/src/config/theme/lyrics.rs
+++ b/src/config/theme/lyrics.rs
@@ -6,19 +6,15 @@ pub struct LyricsConfig {
     pub show_timestamp: bool,
 }
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
 pub struct LyricsConfigFile {
     pub(super) show_timestamp: bool,
 }
 
-impl Default for LyricsConfigFile {
-    fn default() -> Self {
-        Self { show_timestamp: false }
-    }
-}
+impl TryFrom<LyricsConfigFile> for LyricsConfig {
+    type Error = anyhow::Error;
 
-impl LyricsConfigFile {
-    pub(super) fn into_config(self) -> Result<LyricsConfig> {
-        Ok(LyricsConfig { show_timestamp: self.show_timestamp })
+    fn try_from(value: LyricsConfigFile) -> Result<Self> {
+        Ok(LyricsConfig { show_timestamp: value.show_timestamp })
     }
 }
--- a/src/config/theme/lyrics.rs
+++ b/src/config/theme/lyrics.rs
@@ -1,0 +1,24 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Default, Clone)]
+pub struct LyricsConfig {
+    pub show_timestamp: bool,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct LyricsConfigFile {
+    pub(super) show_timestamp: bool,
+}
+
+impl Default for LyricsConfigFile {
+    fn default() -> Self {
+        Self { show_timestamp: false }
+    }
+}
+
+impl LyricsConfigFile {
+    pub(super) fn into_config(self) -> Result<LyricsConfig> {
+        Ok(LyricsConfig { show_timestamp: self.show_timestamp })
+    }
+}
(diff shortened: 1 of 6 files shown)


src/ui/panes/lyrics.rs
```diff
@@ -43,21 +49,34 @@ impl Pane for LyricsPane {
...
-        for line in textwrap::wrap(&current_line.content, area.width as usize) {
+        for (index, l) in
+            textwrap::wrap(&current_line.content, area.width as usize).into_iter().enumerate()
+        {
+            let content = if index == 0 && show_timestamp && !l.is_empty() {
+                format!("[{}] {}", current_line.time.to_string(), l)
+            } else {
+                l.to_string()
+            };
+            let p = Text::from(content).centered().style(middle_style);
```
mierak, line 75: A few issues here. `textwrap::wrap` returns a `Cow<'_, str>` to avoid new String allocations when no wrapping is needed. This implementation forces allocations for every single wrapped line.

You also potentially do some unnecessary work here because you might create the string and then `break` gets called a line after. 

Same issues are in all the other updated loops.

src/config/theme/lyrics.rs
```diff
@@ -0,0 +1,24 @@
...
+pub struct LyricsConfig {
+    pub show_timestamp: bool,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct LyricsConfigFile {
+    pub(super) show_timestamp: bool,
+}
+
+impl Default for LyricsConfigFile {
```
mierak, line 14: You can just derive Default on LyricsConfig file, default for bool is false.

src/config/theme/lyrics.rs
```diff
@@ -0,0 +1,24 @@
...
+    fn default() -> Self {
+        Self { show_timestamp: false }
+    }
+}
+
+impl LyricsConfigFile {
```
mierak, line 20: Use the From (or TryFrom for fallible conversions) trait for simple conversions like this.

src/ui/panes/lyrics.rs
khalil-cheddadi: would this fix the allocation problem?

```rust
let p = if index == 0 && show_timestamp && !l.is_empty() {
   Text::from(format!("[{}] {}", line.time.to_string(), l))
        .centered()
        .style(default_stlye)
} else {
    Text::from(l).centered().style(default_stlye)
};
frame.render_widget(p, *area);
```

mierak: It would, but something like this is better
```rust
let text = Text::from(if index == 0 && show_timestamp && !l.is_empty() {
    Cow::Owned(format!("[{}] {}", current_line.time.to_string(), l))
} else {
    l
})
.centered()
.style(middle_style);
frame.render_widget(text, *area);
```

mierak (commented):
Rebase on top of master and conflict resolution is needed, after that and these comments it should be good to go, thanks!
(and clippy will probably complain too, try running `cargo clippy` after updating to current master)

khalil-cheddadi:
i think i found an issue

this is how wrapped lines normally look:
Screenshot From 2025-06-05 23-56-42

but this is how they look with the added timestamp:
image

i suspect that it is happening because the textwrap did not account for the extra characters

ill try to fix this first

mierak:
Yeah, you're right. But it should be simple to fix by adjusting the area.width by the width of the timestamp in characters (including the one added space) in case timestamp is enabled. Also please use `saturating_sub()` instead of straight `-` here to avoid potential panics.

khalil-cheddadi:
i tried to append the timestamp before passing it to the textwrap
```rs
let formatted_line = if timestamp {
                &format!("[{}] {}", line.time.to_string(), line.content)
            } else {
                &line.content
            };
            for l in textwrap::wrap(formatted_line, area.width as usize).iter().rev() {
           // ...
```

and it worked, but im not sure if i missed something

thank you for being patient with me btw

mierak:
I suppose that is also an option, probably the better one even

> thank you for being patient with me btw

no problem, thank you too for wanting to improve rmpc

src/ui/panes/lyrics.rs
```diff
@@ -43,21 +49,32 @@ impl Pane for LyricsPane {
...
+            &format!("[{}] {}", current_line.time.to_string(), current_line.content)
+        } else {
+            &current_line.content
+        };
+        for l in textwrap::wrap(formatted_line, area.width as usize) {
             let Some(area) = areas.get(current_area) else {
                 break;
             };
-            frame.render_widget(Text::from(line).centered().style(middle_style), *area);
+            let text = Text::from(l.clone()).centered().style(middle_style);
```
mierak, line 76: this clone is completely unnecessary

src/config/theme/lyrics.rs
```diff
@@ -0,0 +1,17 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Default, Clone)]
+pub struct LyricsConfig {
+    pub timestamp: bool,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct LyricsConfigFile {
+    pub(super) timestamp: bool,
```
mierak, line 10: Sorry I just noticed one more thing, can you add the default here?

```suggestion
    #[serde(default)]
    pub(super) timestamp: bool,
```

src/ui/panes/lyrics.rs
```diff
@@ -68,17 +85,21 @@ impl Pane for LyricsPane {
...
+            };
+            for l in textwrap::wrap(formatted_line, area.width as usize).iter().rev() {
                 if before_area_cursor == 0 {
                     break;
                 }
                 let Some(area) = areas.get(before_area_cursor - 1) else {
                     break;
                 };
-                frame.render_widget(p, *area);
+                let text = Text::from(l.clone()).centered().style(default_style);
```
mierak, line 100: clone

src/ui/panes/lyrics.rs
```diff
@@ -93,14 +114,17 @@ impl Pane for LyricsPane {
...
+                &format!("[{}] {}", line.time.to_string(), line.content)
+            } else {
+                &line.content
+            };
+            for l in textwrap::wrap(formatted_line, area.width as usize) {
                 let Some(area) = areas.get(after_area_cursor + 1) else {
                     break;
                 };
-                frame.render_widget(p, *area);
+                let text = Text::from(l.clone()).centered().style(default_style);
```
mierak, line 126: clone

mierak:
thats hopefully the last, I have to go now sorry, I'll take a look in the morning

mierak:
You missed one clone I pointed out above, after that we can merge!

khalil-cheddadi:
evening,
that clone() seems to be necessary because its loop unlike the others has `.iter().rev()`, it was present in the original implementation and removing it causes a type error for `Text::from`

mierak:
Oh, sorry, my bad. That must have slipped during the original PR. I'm merging this then and noting it as a problem for me later. Thanks!

khalil-cheddadi:
thank you and happy happy hacking

v-Nyo:
I think an option to dimm/change the color of the timestamp only would be really nice.

PR #422, opened 2025-06-07
fix: Actually insert and not add playlists when using Insert

tarneaux:
- [x] All tests passed
- [ ] ~Documentation has been updated (if needed)~

There has been no release since the Insert feature was added, making updating `CHANGELOG.md` superfluous.


mierak:
everything looks good to me, thanks!

PR #426, opened 2025-06-08
fix: Actually insert and not add with Insert in search pane

tarneaux:
Somehow this one didn't make it into #422 (fix: Actually insert and not add playlists when using Insert)

Basically just fixing two places where I didn't replace with the correct action when making mpc_client's add commands take a position arg and not an insert one.

### Checklist

- [x] All tests passed

No docs or changelog changes needed.


mierak:
thanks!

PR #428, opened 2025-06-08
fix: correct ordering in Insert and InsertAll on multiple items

### LINKED ISSUE #423: InsertAll reverses order
When using `InsertAll` on selected tracks/playlists/albums, the order of tracks or playlists is reversed since every item is added separately. This could be fixed by modifying the queue position before each add, with a method like this:
```rust
impl QueuePosition {
    ...

    /// Update the position after having added `n` tracks at the same position
    /// (or position passed through this function), to avoid reversing added
    /// lists of tracks.
    #[must_use]
    pub fn update_after_n_adds(&self, n: usize) -> Self {
        use QueuePosition::{Absolute, RelativeAdd, RelativeSub};
        match self {
            RelativeAdd(p) => RelativeAdd(p + n),
            RelativeSub(p) => RelativeSub(*p),
            Absolute(p) => Absolute(p + n),
        }
    }
}
```

I may work on this in the near future, this is just to document the issue.

tarneaux:
Fixes #423 and a similar issue with Insert on a selection.

### DIFF
--- a/src/core/command.rs
+++ b/src/core/command.rs
@@ -213,7 +213,7 @@ impl Command {
                             .collect_vec();
                     }
 
-                    for file in files {
+                    for file in files.into_iter().rev() {
                         if file.starts_with(&dir) {
                             client.add(
                                 file.to_string_lossy()
@@ -231,7 +231,7 @@ impl Command {
                 }))
             }
             Command::Add { files, position, .. } => Ok(Box::new(move |client| {
-                for file in &files {
+                for file in files.into_iter().rev() {
                     client.add(&file.to_string_lossy(), position.clone())?;
                 }
 
--- a/src/ui/panes/playlists.rs
+++ b/src/ui/panes/playlists.rs
@@ -429,7 +429,7 @@ impl BrowserPane<DirOrSong> for PlaylistsPane {
                 });
             }
             [] => {
-                for playlist in &self.stack().current().items {
+                for playlist in self.stack().current().items.iter().rev() {
                     self.add(playlist, context, position.clone())?;
                 }
                 status_info!("All playlists added to queue");
--- a/src/ui/browser.rs
+++ b/src/ui/browser.rs
@@ -363,7 +363,7 @@ where
             }
             CommonAction::AddAllReplace => {}
             CommonAction::Insert if !self.stack().current().marked().is_empty() => {
-                for idx in self.stack().current().marked() {
+                for idx in self.stack().current().marked().iter().rev() {
                     let item = &self.stack().current().items[*idx];
                     self.add(item, context, Some(QueuePosition::RelativeAdd(0)))?;
                 }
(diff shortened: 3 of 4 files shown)


src/core/command.rs
```diff
@@ -231,7 +231,7 @@ impl Command {
                 }))
             }
             Command::Add { files, position, .. } => Ok(Box::new(move |client| {
-                for file in &files {
+                for file in files.into_iter().rev() {
```
mierak, line 234: I think the rev() call here has to be there only if position is given and is of type `RelativeAdd` or `Absolute` (double check me on this :smile:) because like this it will reverse stuff on simple `rmpc add song1 song2 song3` and even on `rmpc add --position -0 song1 song2 song3`

Same with the other Add handler

src/ui/panes/playlists.rs
```diff
@@ -429,7 +429,7 @@ impl BrowserPane<DirOrSong> for PlaylistsPane {
                 });
             }
             [] => {
-                for playlist in &self.stack().current().items {
+                for playlist in self.stack().current().items.iter().rev() {
```
mierak, line 432: And the same probably here. Alternatively the reversing cold be implemented in each pane separately directly in the Insert handler, but thats probably the worse solution.

mierak:
Just throwing ideas, we could also abstract all this directly into the MPD client, create a new add_multiple function which would take list of uris and a position. It could even utilize mpd's command list then. But does not have to be this way for this PR.

src/core/command.rs
doomed-neko: > I think the rev() call here has to be there only if position is given and is of type `RelativeAdd` or `Absolute` (double check me on this 😄) because like this it will reverse stuff on simple `rmpc add song1 song2 song3` and even on `rmpc add --position -0 song1 song2 song3`
> 
> Same with the other Add handler

What's the expected behavior for  `rmpc add --position -0 song1 song2 song3` in this case? Should it insert songs in the same order they were selected in?
I'm kinda interested in implementing this myself but I don't understand what should I aim to do exactly :)

mierak: I think that the position should not influence the order at all. Meaning `rmpc add <position> song1 song2 song3` should result in your queue having song1 song2 and song3 in order no matter what position was given.

doomed-neko: I understand, and happily it seems like the current code works just fine doing that, it worked correctly on all browser panes, the search pane and CLI.
you should probably try it to make sure everything is as it should be :)

mierak:
I think we can close this since it was implemented in https://github.com/mierak/rmpc/pull/468. There is still room for improvement but the core issue is solved. Thanks!

tarneaux:
@doomed-neko Thanks a lot for doing this ! :heart:

doomed-neko:
> @doomed-neko Thanks a lot for doing this ! ❤️

Glad to help :)

PR #436, opened 2025-06-10
feat: Improve cava visualization

polluxcodes:
## Description

This pull request improves the rendering of cava in rmpc using the partially filled unicode blocks (▁▂▃▄▅▆▇█), which creates a smoother-looking visualization.

The `cava.bar_symbol` theme option has been changed to accept multiple characters, which are selected based on how full the grid cell must be.

Example of the new look of Cava:
image

I'm having trouble getting the nightly toolchain working on my computer (Rust and Gentoo sometimes have trouble getting along unfortunately), so I have not yet been able to format the code correctly.

### Checklist

- [x] All tests passed

### DIFF
--- a/src/ui/panes/cava.rs
+++ b/src/ui/panes/cava.rs
@@ -120,8 +120,8 @@ impl CavaPane {
                 if fill_amount < 0.01 {
                     queue!(writer, PrintStyledContent(' '.on(theme.bg_color)))?;
                 } else {
-                    let char_index = (fill_amount * theme.bar_symbol.chars().count() as f32).floor() as usize;
-                    let fill_char = theme.bar_symbol.chars().nth(char_index).unwrap_or(' ');
+                    let char_index = (fill_amount * theme.bar_symbols_count as f32).floor() as usize;
+                    let fill_char = theme.bar_symbols.get(char_index).unwrap_or(&' ');
                     queue!(
                         writer,
                         PrintStyledContent(
--- a/docs/src/content/docs/next/configuration/cava.mdx
+++ b/docs/src/content/docs/next/configuration/cava.mdx
@@ -75,7 +75,7 @@ carefully as the example does not work copy/paste out of the box.
 cava: (
     // symbols that will be used to draw the bar in the visualiser, in ascending order of
     // fill fraction
-    bar_symbol: "▁▂▃▄▅▆▇█",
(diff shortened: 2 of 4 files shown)


src/ui/panes/cava.rs
```diff
@@ -115,15 +116,18 @@ impl CavaPane {
             let color = theme.bar_color.get_color(y as usize, area.height);
             queue!(writer, MoveTo(area.x, h))?;
             for column in columns.iter() {
-                if *column > y {
+                let fill_amount = (*column - f32::from(y)).clamp(0.0, 0.99);
+                if fill_amount < 0.01 {
+                    queue!(writer, PrintStyledContent(' '.on(theme.bg_color)))?;
+                } else {
+                    let char_index = (fill_amount * theme.bar_symbol.chars().count() as f32).floor() as usize;
+                    let fill_char = theme.bar_symbol.chars().nth(char_index).unwrap_or(' ');
```
mierak, line 124: We can store `Vec<char>` in the theme struct instead of String and index into it directly instead of creating chars interator every loop through. Same for the char_index, I would cache the count somewhere (probably in the config struct as well after deserialization, so it gets updated correctly on config reload). Its probably nothing but better to avoid calculations that do not strictly have to be there.

mierak (commented):
Yeah this is sick, thank you! It would have taken me months to get to do this.

mierak:
> I'm having trouble getting the nightly toolchain working on my computer (Rust and Gentoo sometimes have trouble getting along unfortunately), so I have not yet been able to format the code correctly.

I can let this one through and format it myself if its too much trouble.

polluxcodes:
Replaced String with Vec<char> as requested.

I added the field `bar_symbols_count` to the `CavaTheme` struct, which is populated during parsing.

I also took the opportunity to rename `cava.bar_symbol` to `cava.bar_symbols`, since that makes more sense now.

> > I'm having trouble getting the nightly toolchain working on my computer (Rust and Gentoo sometimes have trouble getting along unfortunately), so I have not yet been able to format the code correctly.
> 
> I can let this one through and format it myself if its too much trouble.

Thanks, that would be very helpful!

Should be ready for merging.

src/ui/panes/cava.rs
polluxcodes: Done

mierak:
thank you very much!

PR #446, opened 2025-06-13
feat: remote status timeout

mierak:
## Description

Adds timeout to remote status command. Also store the status messages for future use.

### Checklist

- [x] All tests passed

### DIFF
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -280,6 +280,9 @@ pub enum RemoteCmd {
         #[arg(short, long)]
         #[clap(value_enum, default_value_t = Level::Info)]
         level: Level,
+        /// How long should the message be displayed for in milliseconds
+        #[arg(short, long = "timeout", default_value_t = 5000)]
+        timeout: u64,
     },
     #[clap(hide = true)]
     Tmux { hook: String },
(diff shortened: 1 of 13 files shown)


osamaaj03:
doesn't it make sense to have it permanent when `--timeout` is set to 0, and is there a way to cut the timeout short with a keypress for non-important/low level messages? A quiet option, `-q` or `--quiet`, would be nice too.

mierak:
> doesn't it make sense to have it permanent when --timeout is set to 0

Hm, fair point, though one thing to keep in mind is that the current implementation only looks at the last received status (doesnt matter whether it came from remote or inside) so even an infinite timeout will eventually get replaced.

> and is there a way to cut the timeout short with a keypress for non-important/low level messages

Not really, you can send a remote message with timeout 1 to replace it, but its more of a hack that will eventually stop working than a solution. This whole status system is something that should really get either expanded or replaced but its stuck in a design-stage-limbo.

While Im talking about this, I wanted to have notification popups that slide from the side etc instead of a statusbar. Even have/had a working implementation somewhere, but it got hard blocked by album arts and ratatui not playing along nicely :(.

> A quiet option, -q or --quiet, would be nice too.

What exactly would that option do? If you mean silence the messages printed by `rmpc remote` then sure, that can also be a thing I wouldnt mind. Though you can simply redirect stderr to /dev/null

PR #460, opened 2025-06-16
feat(outputs): Add plugin field to Output struct

johnhamelink:
Add plugin field to track the output plugin type used by MPD audio
outputs. This adds the "plugin" attribute to the output object
generated by `rmpc outputs`.

Example:

```cmdline
john@saturn:~/ > rmpc outputs
[{"id":0,"name":"Snapserver","enabled":true,"plugin":"fifo"}]
```

### DIFF
--- a/src/mpd/commands/outputs.rs
+++ b/src/mpd/commands/outputs.rs
@@ -12,6 +12,7 @@ pub struct Output {
     pub id: u32,
     pub name: String,
     pub enabled: bool,
+    pub plugin: String,
 }
 
 impl FromMpd for Outputs {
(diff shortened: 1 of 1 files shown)


mierak:
Looks good to me, though it would be better to also add it to the `Outputs` modal if you feel up to the task, it should only be a few lines of code as well, though I do not require that.

Other than that, please make a note in the changelog about the change.

mierak:
The PR has been expanded and finished in #465. Thanks!

johnhamelink:
Ah, thanks for finishing it off!

PR #465, opened 2025-06-17
add plugin to outputs (finish #460)

doomed-neko:
## Description

completes #460
I'm having fun doing these tbh, it's like completing quests in a game :)
### Checklist

- [x] All tests passed


doomed-neko:
I am kind of unsure of whether it is a good idea to have `Plugin` take 20% of space and give 80% to `Name`, if you have any ideas I'll be more than happy to hear them

mierak:
I think thats fine. `pipewire` should be the longest plugin name and it can fit just fine with the current settings - the modal is a fixed width.

Thanks!

PR #488, opened 2025-06-24
fix: use 32 bit usize constants in tests

Kladki:
## Description

This is causing rmpc 0.9.0 to fail to compile on Alpine's 32-bit architectures (x86, armv7, armhf). I will probably just add a patch in aports for now so that you don't need to make a new release just for this.

### Checklist

- [x] All tests passed
  - not sure if this is really relevant


mierak:
ah, damn, its always some dumb mistake i make. I suppose I should add a 32bit target to ci test runners Thanks!

I expect more issues to pop up so patch release will most likely happen soon ish.

PR #498, opened 2025-06-24
Feat/add diacritics

### LINKED ISSUE #359: Ability to ignore diacritics when searching/filtering
### Description

In `ncmpcpp` there is an option `ignore_diacritics = yes` and that makes searching easier for me as they are not that uncommon in my music library. Having that option here as well would be a great improvement.

For example searching for `annbjorg` would match `Annbjørg`. Or `lopez` would match both `Lopez` and `López`.

### Proposed solution

_No response_

Gri-ffin:
## Description

I added support for diacritic-insensitive search via a new ignore_diacritics flag on Filter. When enabled, the search input is normalized using Unicode NFD decomposition (via deunicode crate) to remove diacritic marks, this will allow names like lopez to match both Lopez and López, or annbjorg to match Annbjørg I also added additional unit tests to verify that the normalization behaves correctly for a wide range of diacritic use cases, like French, Spanish, German, Norwegian, Swedish, Cyrillic, and CJK characters.

### Related issues

fixes #359 

### Checklist

- [x] All tests passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1954,6 +1954,7 @@ dependencies = [
  "textwrap",
  "thiserror 2.0.12",
  "unicase",
+ "unicode-normalization",
  "unicode-width 0.2.0",
  "url",
  "vergen-gitcl",
@@ -2500,6 +2501,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "tinyvec"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09b3661f17e86524eccd4371ab0429194e0d7c008abb45f7a7495b1719463c71"
+dependencies = [
+ "tinyvec_macros",
+]
+
+[[package]]
+name = "tinyvec_macros"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
+
+[[package]]
 name = "toml"
 version = "0.8.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2550,6 +2566,15 @@ name = "unicode-linebreak"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b09c83c3c29d37506a3e260c08c03743a6bb66a9cd432c6934ab501a190571f"
+
+[[package]]
+name = "unicode-normalization"
+version = "0.1.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5033c97c4262335cded6d6fc3e5c18ab755e1a3dc96376350f3d8e9f009ad956"
+dependencies = [
+ "tinyvec",
+]
 
 [[package]]
 name = "unicode-segmentation"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -586,6 +586,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "deunicode"
+version = "1.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "abd57806937c9cc163efc8ea3910e00a62e2aeb0b8119f1793a978088f8f6b04"
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1930,6 +1936,7 @@ dependencies = [
  "crossbeam",
  "crossterm",
  "derive_more",
+ "deunicode",
  "either",
  "enum-map",
  "flate2",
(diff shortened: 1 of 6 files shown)


mierak (commented):
Hi and thank you.

Few more questions though, how is this supposed to work? Say I have a file where artist is "piñata", if i search for "pinata" in ncmpcpp as any tag, it is found, but nothing is found when I do the same in rmpc with this PR included.

I suspect that is because ncmpcpp lists the whole library and does the search client-side but I have not looked into it too much yet. MPD states in their docs:
```
The find commands are case sensitive, while search and related commands ignore case. The latter also applies [Unicode normalization](https://unicode.org/reports/tr15/) and converts all punctuation to ASCII equivalents if MPD was compiled with [ICU](https://icu.unicode.org/) support.
```
Mine is compiled with ICU but it does not seem to work still, do you know what is going on with that?

Gri-ffin:
Hi, I was doing some tinkering with mpd and its normalization, and i found that while mpd will handle unicode normalization, it will only handle canonical Unicode equivalence (for example i tried a file with Café name, and tried running both ` mpc search title "Café" ` and ` mpc search title "Café" ` and both worked to search for the same song)

But de-diacritization, meaning characters like `'ñ'` are not normalized to `'n'`, and as you said, ncmpcpp handles unicode manipulation on the client side, basically from what i understood it takes the search string, converts it to a `Unicode` string, then applies the ` NFD; [:M:] Remove; NFC ` transliteration rule, and then perform a regex search on this string.

You can look at [regex_filter.h in ncmpcpp](https://github.com/ncmpcpp/ncmpcpp/blob/e8f1c4bd48350789300c5f87c5ea21ea4256cf6f/src/regex_filter.h#L96) to have a better idea.

Gri-ffin:
As for what my code is trying to do, basically let's say i have a file named ` Lonely `, if I try to search for it while typing ` Lonély ` it should work but i think this was a misunderstanding from my part, because for you example if I try a song named ` Lonély ` and i type ` Lonely ` it would not work, since i my code doesn't change `Lonely` to match all the unicodes of e. I was thinking of normalizing the input of the user and sending it to mpd, but I guess that is not what is expected.

So I was thinking maybe there should be a function that will get the user's songs from the mpd server, apply diacritics stripping while iterating through each song on both the song and the user input and match?

mierak:
> But de-diacritization, meaning characters like 'ñ' are not normalized to 'n', and as you said, ncmpcpp handles unicode manipulation on the client side

I think the right call here is asking at MPD's side whether thats expected or not. I would expect 'n' to match 'ñ' there.

> As for what my code is trying to do, basically let's say i have a file named Lonely, if I try to search for it while typing Lonély it should work but i think this was a misunderstanding from my part, because for you example if I try a song named Lonély and i type Lonely it would not work, since i my code doesn't change Lonely to match all the unicodes of e. I was thinking of normalizing the input of the user and sending it to mpd, but I guess that is not what is expected.
>
>So I was thinking maybe there should be a function that will get the user's songs from the mpd server, apply diacritics stripping while iterating through each song on both the song and the user input and match?

I don't have issue with this other than concerns for performance, it is poteintially a lot of data to sift through since you have to fetch all the metadata as well and some users have huge libraries.

Gri-ffin:
>I think the right call here is asking at MPD's side whether thats expected or not. I would expect 'n' to match 'ñ' there.

I searched trough MPD and ICU docs and what i've found is that ICU will only handle cannonical equivalence (like for example 'ñ' ≡ 'n' + ◌̃), and will not handle diacritic stripping because that will break unicode correctness, so MPD will not recognize 'n' as equivalent to 'ñ' and will treat 'ñ' as a different character.

This is mentioned by the ICU docs linked by MPD: [here](https://www.unicode.org/reports/tr15/) and [here](http://www.unicode.org/notes/tn5/#UAX15)

so for diacritic stripping i believe we would have to choose a trade-off:

- Either have performance take the hit and search recursively across the entire directory and subdirectories, strip the chars using icu_normalizer and properties, and try to match just like what `ncmpcpp` does [here](https://github.com/ncmpcpp/ncmpcpp/blob/e8f1c4bd48350789300c5f87c5ea21ea4256cf6f/src/screens/search_engine.cpp)
- Or have the memory take the hit and fetch all songs on startup (and updates) strip them from diacritics and store them in a sort of in memory HashMap or use rusqlite, to match stripped songs with their normalized versions for fast lookup
- Or third option is to raise an issue to mpd itself to implement a native diacritics stripping feature, which would be the most elegant solution, and all clients will benefit from a centralized implementation. But I wouldn't have high hopes since maintainers would have different priorities.

mierak:
I have opened an issue at MPD https://github.com/MusicPlayerDaemon/MPD/issues/2327, it is the solution that makes the most sense to me, but let's see what they have to say.

Otherwise I say we go with option 1, I dont want to maintain copy of MPD's db on client site, that is unnecessarily brittle.

mierak:
I attempted to implement this in MPD, lets see where it goes https://github.com/MusicPlayerDaemon/MPD/pull/2329

mierak:
The PR at MPD got finally merged, I have incorporated your changes and expanded upon them in https://github.com/mierak/rmpc/pull/575 so I am closing this. Thanks!

PR #515, opened 2025-07-01
docs: fix typos

AGN907:
## Description
Browsing the keybinding and cava page to configure rmpc and finding some typos.

I only fixed the next release docs, tell me if you want me to also change the current release docs
### Checklist

- [ ] All tests passed


mierak:
Appreciate it! I would be fine with just the next section, but I apparently fumbled a config option in cava so that would be nice to get fixed in the release too so users dont copy a wrong value.

mierak:
thanks!

PR #516, opened 2025-07-01
fix/standard time format

### LINKED ISSUE #490: Consistent duration formatting
### Description

With the introduction of the `QueueTimeTotal` and `QueueTimeRemaining` properties, there are now two formats of durations. There is the format used by the above properties, and then there is the format used, to my knowledge, everywhere else.

Now, IMO, the format for the `QueueTime` properties is a bit odd. It's not really a standard way of formatting times, and only being able to change the separator doesn't do much to change that. I would like to be able to at least change it to be consistent with the way other durations are formatted.

### Proposed solution

Ideally, there would be a single duration formatting function, which you could give a conventional format string, e.g., `"%Dd %H:%M:%S"` to get times formatted as, e.g., `3d 15:20:13`.

Now, I'm not too familiar with Rust and its formatting facilities, but from what I was able to gather, this would require implementing quite a bit of logic, especially if you wanted to omit zero values (e.g., showing `45:12` in the above format if the duration was under an hour). However, it would facilitate any format you want to give the duration. But, it's also a bit of an overkill considering time formatting is pretty standardized.

A simpler solution that would enable consistent formatting while also retaining the current formatting of the `QueueTime` properties for those who prefer it would be to 1. allow passing `separator: None` to the properties, in which case their formatting would fall back to the `to_string` method, and 2. expand the `to_string` method on durations to something like
```rust
fn to_string(&self) -> String {
    let secs = self.as_secs();
    
    let min = secs / 60;
    let frac_secs = secs - min * 60;

    let hours = min / 60;
    let frac_min = min - hours * 60;

    let days = hours / 24;
    let frac_hours = hours - days * 24;

    if hours == 0 {
        format!("{}:{:0>2}", min, frac_secs);
    } else if days == 0 {
        format!("{}:{:0>2}:{:0>2}", hours, frac_min, frac_secs);
    } else {
        format!("{}d {:0>2}:{:0>2}:{:0>2}", days, frac_hours, frac_min, frac_secs);
    }
}
```

Gri-ffin:
## Description

This PR implements consistent duration formatting across all time-based properties in rmpc, now `QueueTimeRemaining` and `QueueTimeTotal` will default to `":"` unless user specifies a separator, to keep it consistent formatting with `Elapsed` and `Duration`, I also added support for hours and days just in case.

### Related issues

closes #490

### Checklist

- [x] All tests passed

### DIFF
--- a/src/shared/ext.rs
+++ b/src/shared/ext.rs
@@ -218,7 +218,7 @@ pub mod duration {
     impl DurationExt for std::time::Duration {
         fn to_string(&self) -> String {
             let secs = self.as_secs();
-            
+
             let min = secs / 60;
             let frac_secs = secs - min * 60;
 
--- a/src/ui/panes/mod.rs
+++ b/src/ui/panes/mod.rs
@@ -1590,7 +1590,7 @@ mod format_tests {
         // With elapsed time, remaining should subtract elapsed from current song
         #[case(StatusProperty::QueueTimeRemaining { separator: None }, "5:49", Duration::from_secs(20))]
         #[case(StatusProperty::QueueTimeRemaining { separator: None }, "5:09", Duration::from_secs(60))]
-        // Verbose format tests (with separator = verbose format)  
+        // Verbose format tests (with separator = verbose format)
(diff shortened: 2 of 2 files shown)


mierak:
This looks very nice, thank you, especially for the tests since the property ones are a pain to write.

Issue #400, opened 2025-06-02
No autoplay


roda37:
### Description

Is there any option where i can toggle autoplay? As much as I understand, you are supposed to add things to playlist and then play it as is. This is painful, when  I want music, I turn it on and it plays until i stop it, burning through that directory. Is there any way to do this because I couldn't find it in docs.

### Steps to reproduce

a

### rmpc debuginfo

<details>
  <summary>rmpc debuginfo</summary>

  ```sh
Paste the output here
  ```
</details>

### Config file

<details>
  <summary>rmpc config</summary>

  ```ron
Paste the output here
  ```
</details>

### Relevant log output

<details>
  <summary>rmpc log</summary>

  ```
Paste the contents here
  ```
</details>

mierak:
If you mean that you want to play the whole directory without adding it to the queue then no, that is simply not possible, MPD does not work that way.

If you want to add something to the queue and play it right away at the same time, then this could be added in some way, shape or form. There is also the confirm action which either opens a directory or adds and plays a single song, not a whole directory though.

jgregoire:
A feature I'm used to in other clients is "Add and play" where it immediately starts playing whatever was just added. I haven't seen a keybind for that.

[edit] I know I can do this with individual songs with <CR> by default, but I'd like to be able to do it with whole albums, artists, or playlists, too.

joereynolds:
Commenting to purely say I'm interested in this too.

Atm to "play an album" I have to go through each song, add it to the queue and then play.

Would be good if there was a way to add an album and start playing

mierak:
You can add the whole album by pressing the `Add` keybind on the "directory" that already works. You can also go to search, search for the album and press `AddAll`. Though you still have to go to the queue and start playing them.

Anyway, this is on the roadmap.

mierak:
Since this is getting some interest, doesn't hurt to ask for opinions. I am thinking of several possible solutions:

* A new keybind, something like `AddAndPlay`, this then gets into things like do we need `AddAllAndPlay`? Do we also need `InsertAndPlay` to add after the current song etc. and gets messy real quick
* A new parameterized keybind `AddAndPlay(position: EndOfPlaylist | AfterCurrentSong | ..., single: true)` - this gets pretty messy pretty quick as well
* Take inspiration form ncmpcpp, it has this nifty little modal flow which could be implemented and gives great flexibility and extendability for the future but is multiple keystrokes for something that could have been a keybind
[...]

Issue #276, opened 2025-03-12
Rounding options for borders


acdcbyl:
### Description

Hello! After the author's efforts to update rmpc has become a very elegant and comfortable terminal player, I like it too much ♥️! The downside is that there doesn't seem to be a way to round the corners (maybe I'm being a bit fussy - straight corners just don't seem to go well with my terminal). I'm looking forward to being able to add an option for borders if that's convenient. Finally, I hope everything goes well for you!

![Image](https://github.com/user-attachments/assets/b3612448-6d54-43f3-b219-e918be19916d)

mierak:
This was already sort of requested and I talked about it here https://github.com/mierak/rmpc/pull/239#issuecomment-2654060281. But TLDR is, I want to implement that too just unsure about the details.

mierak:
Finally possible to do this and more #793!
https://rmpc.mierak.dev/next/configuration/tabs/#border_symbols

PR #518, opened 2025-07-02
feat: better add to queue

mierak:
## Description

Work in progress. Comes with a new menu modal and a configurable keybind 

This PR also changes the behavior of `Confirm` action in the browser. It no longer adds the single song to end of queue and plays it but instead replaces the whole queue with the contents of the current directory and plays the selected item.

Some examples, without any kind specified it opens modal with default settings. The modal is also configurable. Everything should also handle single and multiple marked items.
```ron
"<C-a>":       AddOptions(),
"<C-b>":       AddOptions(kind: Modal([
                   ("End of queue and play", (autoplay: Hovered, all: false, position: EndOfQueue)),
                   ("Replace queue and play", (autoplay: Hovered, all: false, position: Replace)),
               ])),
"<C-q>":       AddOptions(kind: Action(autoplay: None, all: false, position: StartOfQueue)),
"<C-w>":       AddOptions(kind: Action(autoplay: None, all: false, position: EndOfQueue)),
"<C-f>":       AddOptions(kind: Action(autoplay: None, all: false, position: BeforeCurrentSong)),
"<C-p>":       AddOptions(kind: Action(autoplay: None, all: false, position: AfterCurrentSong)),
```

The current `Add`, `Insert`, `AddAndReplace`, `AddAll`, `InsertAll` and `AddAllReplace` actions have been removed from the code but are kept in the config file and mapped to this new generic one for compatibility.

image

### Related issues

https://github.com/mierak/rmpc/issues/400

### Checklist

- [ ] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,17 @@ All notable changes to this project will be documented in this file.
 - Added `Transform` properties and `Truncate` transformation
 - Added `sendmessage` cli command for inter client communication
 - Support for youtube playlists
+- Added configurable kebyind for adding items to the Queue - `AddOptions`
 
 ### Changed
 
 - Normalized duration formatting across `QueueTimeTotal` and `QueueTimeRemaining` properties for both standard (MM:SS/H:MM:SS) and verbose formats
+- Confirm action in browsers, which either opens a directory or adds the hovered song to the queue
+no longer exhibits the latter behavior. It now instead replaces the queue with all songs in the
+directory.
+- Removed `Add`, `Insert`, `AddAndReplace`, `AddAll`, `InsertAll` and `AddAllReplace` from the code.
+They are now mapped to the new `AddOptions` action. Existing configs are not affected and will
+continue to work.
 
 ### Fixed
 
--- a/docs/src/content/docs/next/configuration/keybinds.mdx
+++ b/docs/src/content/docs/next/configuration/keybinds.mdx
@@ -132,11 +132,77 @@ list and controlling search mode.
 |       `r`       | Rename          | Rename. Currently only for playlists                                                                                               |
 |       `a`       | Add             | Add item to queue                                                                                                                  |
 |       `A`       | AddAll          | Add all items to queue                                                                                                             |
-|                 | AddReplace      | Replace current queue with the item                                                                                                |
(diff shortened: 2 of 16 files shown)


Solarunit:
would it be possible to edit this menu, e.g. remove or move items on top. For example I want `Replace queue and play` be on top and remove some items I never use.

On the right side in the menu there could be shortcuts visible (using 60% transparent font)

mierak:
> would it be possible to edit this menu, e.g. remove or move items on top. For example I want Replace queue and play be on top and remove some items I never use.

yup, sounds good

> On the right side in the menu there could be shortcuts visible (using 60% transparent font)

Maybe, theres no such a thing as a transparent font in the terminal though. The closest you can get is synthesize a color somewhere between the fg and bg and I dont really wanna do that because that really only works for rgb and hex colors.

sabisho:
Will transparency work on this floating box?
Also making the menu box customizable would be better in the long term.

mierak:
> Will transparency work on this floating box?

I mean, if your terminal is transparent and you set no bg color in rmpc, then rmpc will be transparent as well, including this and any other modals, there are no issues with that and this PR does not change anything about it.
The thing I am talking about is that I cannot easily take the foreground color of your terminal and make it 60% opacity, that does not exist.

> Also making the menu box customizable would be better in the long term.

Yup, agree.

sabisho:
> I mean, if your terminal is transparent and you set no bg color in rmpc, then rmpc will be transparent as well, including this and any other modals, there are no issues with that and this PR does not change anything about it. The thing I am talking about is that I cannot easily take the foreground color of your terminal and make it 60% opacity, that does not exist.
Yes I have transparency enabled.

BTW, how do I enable border rounding of different panes, like you did in this menu box?

mierak:
> BTW, how do I enable border rounding of different panes, like you did in this menu box?

https://github.com/mierak/rmpc/issues/276 on hold untill the next ratatui version. That version already has two relevant things merged and ready, one of which was requested by me. Once that is out the borders can be improved.

mierak:
There is still a bunch of work left to do on this, but it is very, very slowly getting to a usable state. So if you wanna have some early feedback have a go at it. Examples are in the PR description.

I am also looking for alternatives to `AddOptions` since I am not in love with the name if you have any ideas.

PR #520, opened 2025-07-04
feat/improve LRC parsing logic

### LINKED ISSUE #519: Lyrics parsing doesn't allow brackets in text
### Description

Me again, with an issue related to #392 :
In lrc files like this:
```lrc
[title: Song Name [Explicit]]
...
[00:09.00]
[00:10.00] [Drum Solo]
[00:11.00]
```
the tag will be parsed correctly, but the `[Drum Solo]` will not.

I almost wanted to start improving the code, when I remembered that libraries exist, so maybe using https://crates.io/crates/lrc is the easiest solution.

I might look at it in the next few days, so I can finally enjoy my weirdly bracketful lrc files :)

### Steps to reproduce

Put any word in one of your lrc files in brackets

### Expected behavior

The bracketed word should be parsed as part of the lyrics and not the timestamps

### What terminal emulator are you running rmpc in?

kitty 0.42.1

### rmpc debuginfo

<details>
  <summary>rmpc debuginfo</summary>

  ```sh
rmpc 0.9.0 git Arch Linux

Config path          /home/lin/.config/rmpc/config.ron
Theme path           Some("lxn")

MPD:
Address              "127.0.0.1:6600"
Resolved Address     IpAndPort("127.0.0.1:6600")
MPD_HOST             unset
MPD_PORT             unset

Youtube playback:
Cache dir            None
ffmpeg               installed       ffmpeg version n7.1.1 Copyright (c) 2000-2025 the FFmpeg developers
ffprobe              installed       ffprobe version n7.1.1 Copyright (c) 2007-2025 the FFmpeg developers
yt-dlp               installed       2025.06.30          
python3              installed       Python 3.13.5       
python-mutagen       not installed   NOT PRESENT         

Image protocol:
Requested            Auto
Resolved             Kitty
TMUX                 false
ueberzugpp           not installed   Version not available

Visualizer:
cava                 not installed   Version not available
  ```
</details>

### Config file

<details>
  <summary>rmpc config</summary>

  ```ron
Paste the output here
  ```
</details>

### Relevant log output

<details>
  <summary>rmpc log</summary>

  ```
Paste the contents here
  ```
</details>

Gri-ffin:
## Description

This PR introduces a significant change in how the LRC parser handles errors and bracketed text inside lyrics.

**Key changes:**

* I made the parser more forgiving by having it skip over malformed lines, invalid timestamps, or bad metadata instead of failing or crashing. Since real-world `.lrc` files can be messy or inconsistent, I believe it’s better to show partial lyrics than none at all for a better ux, plus how its mentioned in the issue itself.

* I also fixed how the parser handles brackets inside lyrics. Previously, bracketed text like `[Drum Solo]` was mistakenly treated as metadata tags, causing those parts to be lost or leading to parsing failures. Now, the parser correctly distinguishes between actual tags (timestamps and metadata) and brackets that are part of the lyrics, and skip over comments.

* For more info you can look at the tests to see how the parser should behave like.

Since this changes fundamental parsing behavior, I’m looking for feedback on whether this forgiving approach aligns with the expectations or if a stricter error policy would be better.

I haven’t updated the docs or changelog yet, as I wanted to agree on the overall approach before proceeding.

### Related issues

fixes #519

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-
 - Improved `.lrc` lyrics files parser performance and fixed parsing issues
 - Normalized duration formatting across `QueueTimeTotal` and `QueueTimeRemaining` properties for both standard (MM:SS/H:MM:SS) and verbose formats
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+
+- Improved `.lrc` lyrics files parser performance and fixed parsing issues
 - Normalized duration formatting across `QueueTimeTotal` and `QueueTimeRemaining` properties for both standard (MM:SS/H:MM:SS) and verbose formats
 
 ### Fixed
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -188,7 +188,7 @@ impl LrcIndex {
 impl LrcIndexEntry {
     fn read(mut read: impl BufRead, path: PathBuf) -> Result<Option<Self>> {
         let mut content = String::new();
-        
+
         loop {
             let mut line = String::new();
             if read.read_line(&mut line)? == 0 {
@@ -201,7 +201,9 @@ impl LrcIndexEntry {
             if !trimmed.is_empty() && !trimmed.starts_with('#') && trimmed.starts_with('[') {
                 if let Some(bracket_end) = trimmed.find(']') {
                     let tag_content = &trimmed[1..bracket_end];
-                    if tag_content.chars().next().is_some_and(|c| c.is_numeric()) && tag_content.contains(':') {
+                    if tag_content.chars().next().is_some_and(|c| c.is_numeric())
+                        && tag_content.contains(':')
+                    {
                         // timestamp found, add this line and stop
                         content.push_str(&line);
                         break;
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -4,13 +4,13 @@ use std::{
     time::Duration,
 };
 
-use anyhow::{Context, Result, bail};
+use anyhow::{Context, Result};
 use itertools::Itertools;
 use serde::Serialize;
 use unicase::UniCase;
 use walkdir::WalkDir;
 
-use super::{Lrc, parse_length};
+use super::{Lrc, parse_metadata_only};
 use crate::{mpd::commands::Song, shared::macros::try_cont};
 
 #[derive(Debug, Eq, PartialEq, Default, Serialize)]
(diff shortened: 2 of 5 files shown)


mierak:
You beat me to the punch, I started implementing the same approach too but didnt get this far. Thanks!

I agree with pretty much everything you said. One thing to note though is the index file in the lyrics. It is used for building the lrc index on startup and should reuse this logic as well so that the metadata is parsed consistently.
It should also stop parsing after the metadata is done when the actual lyrics start so no data is read unnecessarily on start.

I am still looking at the tests, but it seems to behave great.

Gri-ffin:
yeah thanks, i'll move the metadata handling to a separate function then and make the main parser only handle offset and timestamps.

Gri-ffin:
for this new code i thought since we already parse metadata we shouldn't re-read it again, we should just return the index where the timestamp starts, although metadata wouldn't be too big but it would increase performance (technically).

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+
+- Improved `.lrc` lyrics files parser performance and fixed parsing issues
```
Alexdelia, line 17: _nitpick_: looks like an extra new line
```suggestion
- Improved `.lrc` lyrics files parser performance and fixed parsing issues
```

mierak:
> for this new code i thought since we already parse metadata we shouldn't re-read it again, we should just return the index where the timestamp starts, although metadata wouldn't be too big but it would increase performance (technically).

Smart, but I'm not sure its worth the effort. I am more worried about performance of the indexing itself since that reads many files at once. Reading a single file on song change is nowhere near as a hot path. It would also open up some TOCTOU issues if the lrc file changed between indexing actually reading.

src/shared/lrc/index.rs
```diff
@@ -175,57 +186,43 @@ impl LrcIndex {
...
-                        "ti" => title = Some(value.trim().to_string()),
-                        "ar" => artist = Some(value.trim().to_string()),
-                        "al" => album = Some(value.trim().to_string()),
-                        "length" => length = Some(parse_length(value.trim())?),
-                        _ => {}
+            // if this line has a timestamp, stop reading
+            // We are looking for lines that start with [ and have a timestamp in them
+            // reading all the way to the end of the file is not necessary
+            let trimmed = line.trim();
+            if !trimmed.is_empty() && !trimmed.starts_with('#') && trimmed.starts_with('[') {
```
mierak, line 201:
```suggestion
            if !trimmed.is_empty() && trimmed.starts_with('[') 
```
This check is redundant right?

src/shared/lrc/index.rs
```diff
@@ -175,57 +186,43 @@ impl LrcIndex {
...
 
-            let (metadata, rest) = buf
-                .trim()
-                .strip_prefix('[')
-                .and_then(|s| s.rsplit_once(']'))
-                .with_context(|| format!("Invalid lrc line format: '{buf}'"))?;
-            if !rest.is_empty() {
-                break;
+        loop {
+            let mut line = String::new();
```
mierak, line 193: This buffer can be reused by moving it just above the loop and clearing at the end of each iteration. Saves a few reallocations.

src/shared/lrc/lyrics.rs
```diff
@@ -1,114 +1,279 @@
...
-                        "length" => result.length = Some(parse_length(value.trim())?),
-                        _ => {}
+                let Some(close_pos) = close_pos else {
+                    break; // No closing bracket found
+                };
+                let tag_content = &remaining[..close_pos];
+                let is_timestamp = tag_content.chars().next().is_some_and(|c| c.is_numeric())
+                    && tag_content.contains(':');
+                let is_metadata = !is_timestamp && tag_content.contains(':');
+                if is_timestamp || is_metadata {
```
mierak, line 198:
```suggestion
                if is_timestamp && !is_metadata {
```

There shouldn't be a reason to push metadata to the tags vec at this point right? It gets ignored after.
Maybe we should also mention that metadata tags are ignored inside the actual lyrics in the file.

mierak: Maybe not though, the current implementation also supports things like 
`[00:10.00][str:rst][00:12.00]repeated with mangled tag in the middle` which results in two repeated lyrics lines. The suggested change would break it.

But then again `[00:10.00][str][00:12.00]repeated with mangled tag in the middle` (without `:`) in the unexpected tag) already doesnt work with current impl. It resolves to a single line with `[str][00:12.00]repeated with mangled tag in the middle` as content.

mierak:
Other than these few minor things I see no issues, great work, thanks!

src/shared/lrc/lyrics.rs
Gri-ffin: You're right, thanks for pointing that out, now I changed the behavior to be more conservative. The parser now **only processes consecutive timestamps at the start of a line**, stopping at the first non-timestamp tag (whether it's a valid metadata tag like `[str:rst]` or an invalid one like `[str]`).

For example:

```
[00:10.00][str:rst][00:12.00]lyrics
```

This creates a **single lyric line at `00:10.00`** with the content:

```
[str:rst][00:12.00]lyrics
```

It **does not attempt to parse** the later timestamp `[00:12.00]`.

While this might seem less "smart" when dealing with malformed files, i think it's a **safer choice**. It preserves expected behavior in normal cases like:

```
[00:10.00][Intro] Welcome
```

Where bracketed text like `[Intro]` should be treated as part of the lyrics, not misinterpreted as a timestamp or metadata.

If we try to implement something that handle those case we'll end up with a parser using complex heuristics to distinguish between metadata, invalid tags, and legitimate bracketed, which would be hard, time consuming, ambiguous and error-prone. This will also users to spot and clean up malformed timestamps in their files themselves when they are rendered.

src/shared/lrc/index.rs
Gri-ffin: yes, i forgot to change after refractoring the code

mierak:
Just formatting in the latest commits, otherwise its good, thanks!

Lxtharia:
Thank you @Gri-ffin!

PR #533, opened 2025-07-07
feat/command_keybind_cli_commands

### LINKED ISSUE #499: Emulate keybind press or command execution from CLI
### Description

I want to switch to lyrics tab (and back to browse tab) quickly without having to navigate to rmpc instance on my second monitor.

With my proposed solution rmpc cli command could be binded in window manager config.

### Proposed solution

```bash
rmpc remote keybind "7" # "7" is keybind for lyrics tab
```

or

```bash
rmpc remote command "SwitchToTab("Lyrics")"
```

On top of that a way to check current tab would allow to make a simple switch.

`if current_tab = browser then switch-tab(lyrics) else switch-tab(browse)`

Gri-ffin:
## Description

This PR introduces new remote subcommands for IPC control:

`rmpc remote keybind <key>` and `rmpc remote command <action>`.

While reading issue #499, I noticed how these remote command features could greatly benefit workflows with window managers and external tools, so I decided to tackle the implementation myself for the past few days.

At first I didn't know which direction to go with, either 
```
rmpc remote keybind "7" # "7" is keybind for lyrics tab
```
or 
```
rmpc remote command "SwitchToTab("Lyrics")"
```
So I ended up implementing them both to allow the user the freedom of choice.

### Related issues

fixes #499 

### Checklist

- [x] All tests passed


mierak:
Sick, thanks, I will have a closer look tomorrow because I have to leave atm, but some thoughts for now:

`rmpc remote keybind ...` looks great!

I am not sure that executing CLI commands remotely has much benefit? By that I mean - why would I execute `rmpc remote command play` instead of just `rmpc play`? - the IPC is strictly local so both of these rmpc instances have access to the same MPD instance.
It is also a bit confusing that `rmpc remote command` special cases a singular keybind (SwitchToTab) otherwise parses them as CLI. The way I understood the issue was being able to execute the kebyind actions, meaning manipulating rmpc, not mpd, which makes more sense to me.

I was thinking about this issue as well - especially with the point I made about the stupidly namespaced keybinds. Maybe it would be better to introduce a separate set of actions that can be executed remotely and curate them one by one as needed? Say we start with `switch-tab "Queue"` (not sure about the casing, whether to keep it the same as the current Keybinds etc, but you get the idea). Others I can see could be `switch-partition <name>` as I am working on that issue right now and other things that manipulate rmpc. It would then beg a question whether we would need `rmpc remote command switch-tab` or just `rmpc remote switch-tab`. What do you think?

Gri-ffin:
Hi,

My initial thinking on that rmpc remote command is that I was exploring two paths. On one hand, simulating keybinds (which led to `rmpc remote keybind`). On the other, I was trying to find a way to directly invoke rmpc's internal actions. `rmpc remote command "SwitchToTab(\"Lyrics\")"` felt like a more explicit, perhaps less "hacky," way to control the UI than just relying on a numeric keybind. My hope was that it would offer a really precise, explicit way for external tools like window managers or scripts to manipulate rmpc's state, rather than just proxying MPD commands. But, as you pointed out, trying to lump both generic command proxying and specific rmpc client actions under that single, broad rmpc remote command umbrella definitely created inconsistency and blurred the lines between different concerns. It wasn't quite hitting the mark that I intended.

So, maybe the best way to go forward now is the make this PR to focus solely on the rmpc remote keybind subcommand, this would be much cleaner, make the PR focus on one thing and makes it simpler to review, I can leave the `remote command` until you settle on what it should do specifically and what it should look like and have it's own distinct PR for it.

Gri-ffin:
Also I like your approach of starting with specific, curated commands rather than trying to expose everything at once. It's much cleaner to add switch-tab and switch-partition as actual use cases come up, instead of building a generic command parser.

The rmpc remote keybind would already handles the "trigger any action" case anyway, so future remote commands can just focus on making the common stuff easier to use. Also I think we should go with `rmpc remote switch-tab`, it follows common CLI patterns:

- git remote add
- docker container stop
- systemctl user enable

having `command` would just be another word to type, and `rmpc remote --help` would show all the available actions directly, making it more discoverable than hiding them behind a generic "command" subcommand.

If so should i try to implement it along with this PR or maybe leave this PR for the `keybind` command specifically and push another one for `switch-tab`?

mierak:
> So, maybe the best way to go forward now is the make this PR to focus solely on the rmpc remote keybind subcommand

> Also I think we should go with rmpc remote switch-tab, it follows common CLI patterns

Yup, I agree. I gave it a bit more thought and I do like the approach.


> If so should i try to implement it along with this PR or maybe leave this PR for the keybind command specifically and push another one for switch-tab?

I think I will leave that up to you. I would do it as a separate PR personally but dont mind either way.

And thanks again, you have been a huge help.

mierak:
looks pretty good to me! thanks

PR #536, opened 2025-07-09
feat/switch-tab-command

### LINKED ISSUE #499: Emulate keybind press or command execution from CLI

Gri-ffin:
## Description

This PR is a follow up on the discussion in #533, it introduces the first new remote command which allows to switch tabs.

Examples:

```
# Switch to Queue tab
rmpc remote switch-tab Queue

# Case-insensitive matching works
rmpc remote switch-tab queue
rmpc remote switch-tab ARTISTS
rmpc remote switch-tab browser

# Invalid tab names are logged, not displayed to user
rmpc remote switch-tab InvalidTab  # Logs error internally, returns failure code
```

I made tab name matching case-insensitive to enhance the user experience, my thinking is that users should not type `Queue` as it would be cumbersome, or they might not remember the exact capitalization. 

For error handling, invalid tab names are logged internally internally rather than displaying to the user an error message, current implementation somehow seems like a "current fire-and-forget" socket communication design where commands are sent without waiting for execution results. 

I could've have "hacked" this by pre-validating tab names in the CLI using local config, but this would be inconsistent with other remote commands and can lead to race conditions if tab configuration changes between CLI validation and command execution. It seems to me like the proper solution would be to modify the socket protocol to support request-response communication, but that felt like a design decision that should be left to @mierak.

### Related issues

fixes #499

### Checklist

- [x] All tests passed

### DIFF
--- a/src/shared/ipc/commands/switch_tab.rs
+++ b/src/shared/ipc/commands/switch_tab.rs
@@ -22,30 +22,22 @@ impl SocketCommandExecute for SwitchTabCommand {
         _work_tx: &Sender<WorkRequest>,
         config: &Config,
     ) -> Result<()> {
-        // tabs are case-insensitive matching
-        let target_tab_name = config
+        let tab_name = config
             .tabs
             .names
             .iter()
-            .find(|name| name.to_lowercase() == self.tab.to_lowercase())
-            .cloned();
-
-        let tab_name = match target_tab_name {
-            Some(name) => name,
-            None => {
-                return Err(anyhow::anyhow!(
-                    "Tab '{}' does not exist in configuration. Available tabs: {}",
-                    self.tab,
-                    config
-                        .tabs
-                        .names
-                        .iter()
-                        .map(|name| name.to_string())
-                        .collect::<Vec<_>>()
-                        .join(", ")
-                ));
-            }
-        };
+            .find(|name| name.as_str().eq_ignore_ascii_case(&self.tab))
+            .cloned()
+            .ok_or_else(|| {
+                let available = config.tabs.names.iter()
+                    .map(|name| name.as_str())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                anyhow::anyhow!(
+                    "Tab '{}' does not exist. Available tabs: {}", 
+                    self.tab, available
+                )
+            })?;
 
         event_tx.send(AppEvent::UiEvent(UiAppEvent::ChangeTab(tab_name)))?;
         Ok(())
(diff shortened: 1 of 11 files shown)


src/shared/ipc/commands/switch_tab.rs
```diff
@@ -0,0 +1,45 @@
...
+            .ok_or_else(|| {
+                let available = config
+                    .tabs
+                    .names
+                    .iter()
+                    .map(|name| name.as_str())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                anyhow::anyhow!("Tab '{}' does not exist. Available tabs: {}", self.tab, available)
+            })?;
```
mierak, line 40: I would skip the validation here altogether or move it inside the event loop, the main reason being config hot reload.

The hot reloading only changes config in the main thread, there is currently no mechanism to notify other threads about the change and thus this config is not guaranteed to have up to date tab config.

src/shared/ipc/commands/switch_tab.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        event_tx.send(AppEvent::UiEvent(UiAppEvent::ChangeTab(tab_name)))?;
```
mierak, line 42: For the same reason as above, I think its better to create a new AppEvent, handle it in event loop, do the validation there (or just let the `change_tab` fail, whichever is better) and then fire the ChangeTab event from the loop.

I would also display a `status_error!` if that validation fails.

mierak (commented):
> I made tab name matching case-insensitive to enhance the user experience, my thinking is that users should not type Queue as it would be cumbersome, or they might not remember the exact capitalization.

I agree, but this should probably be done by implementing a custom PartialEq and Eq on the `TabName` struct instead of the derived ones (using the UniCase crate, its already used in a few places). That way it will be consistent between config and the remote command. This should be noted as a breaking change even though it most likely won't affect anybody.

> For error handling, invalid tab names are logged internally internally rather than displaying to the user an error message, current implementation somehow seems like a "fire-and-forget" socket communication design where commands are sent without waiting for execution results.

My rationale here was that the `rmpc remote` will fail if the sending itself does not succeed, but if some other logic error occurs it should get either logged or displayed as a status in the running rmpc instance.

> I could've have "hacked" this by pre-validating tab names in the CLI using local config

I think just displaying the error status in rmpc's tui as I outlined in the comments below is fine for this PR. But we will have to somehow make the communication two way anyway when we get to the query IPC.

mierak:
Just some rogue formatting issues, otherwise looks good!

mierak:
thanks!

Issue #491, opened 2025-06-24
Context menus support


Solarunit:
### Description

I think having context menus (right click menu) everywhere in the "UI" would improve usability significantly.
For example in Playlists tab you could right click on any playlists and see menu with:

 - add to the Queue
 - clean Queue and add
 - remove
 - rename
 - etc..

or in the Queue tab, right click on any song would bring context menu with:

- track info
- delete track from playlist
- etc..

____

Context menus could be triggered with right mouse click or some shortcut.

mierak:
very similar, but not exactly equivalent to https://github.com/mierak/rmpc/issues/400 it makes sense to me to have something like this as well

Solarunit:
it also would be very useful for Dirs Tab.
Context menu for selected directory to add it to saved playlist or create a new one.

Also for Playlists tab:

- create new playlist
- delete playlist
- rename playlist

Issue #531, opened 2025-07-07
Modal window to add directories to the Playlists


Solarunit:
### Description

Similar to the new `AddOptions` modal window, there should be one for Directories Tab for adding directories to the saved playlists or create new playlist when adding a folder.

Or such items could be combined into single modal window (with a line separator). For example item `Add folder to the playlist` should be only visible in Directories Tab.

In the future this modal window could be triggered with right mouse click.

Issue #493, opened 2025-06-24
Auto name new Playlist when adding folder to it


Solarunit:
### Description

I would like to see this great usability feature to automatically name new Playlist when you add music folder(s) to it, like in some other music players. 

- the name should be created only when you create playlist for the 1st time
- name for the playlist should be Artist, Artist-album, or an adding folder name (if there are many artists in the album/folder)

For example, if you add folder Coldplay with many sub-folders, playlist should be named as "Coldplay"

____
because renaming playlists is a pain.

PR #539, opened 2025-07-12
feat: context menu

mierak:
## Description

A bit larger PR than I would have hoped due to some needed refactors/small fixes.

Adds a new `ContextMenu()` action. The empty parens are needed to keep options open for some configuration
in the future.
This opens the new menu modal with select few options that can be done on selected item(s). The context menu can also be opened with right click.

The `shared::ext` module was getting quite large so mpd client extensions have been moved to a separate file.

This PR also fixes `ModalClosed` event to only be dispatched when all modals are closed, not any,
which mean we can now safely open modal from within a modal as is done in the context menu and also
fixes a few small issues around album art/cava.
`pop_modal!()` macro was also removed and each modal instance now has a `hide()` function instead.

### Related issues

fixes https://github.com/mierak/rmpc/issues/491
partially addresses https://github.com/mierak/rmpc/issues/493 and https://github.com/mierak/rmpc/issues/531

### Checklist

- [x] All tests passed


Solarunit:
I tried it, nice!

1. is there a a way to highlight item on cursor hover and activate it on a single click?
2. So now we have modal windows and context menus with slightly different options, which is confusing. You have to remember where that one particular option is.
3. Is there a default shortcut to open context menu?

mierak:
> is there a a way to highlight item on cursor hover and activate it on a single click?

no

> Is there a default shortcut to open context menu?

`<C-z>`

> So now we have modal windows and context menus with slightly different options, which is confusing. You have to remember where that one particular option is.

not sure where you are going with this, you are the one who requested this exact thing

Solarunit:
>not sure where you are going with this, you are the one who requested this exact thing

so in the Playlists Tab we have context menu with:

- Add to queue
- **Replace queue**
- Create playlist
- Add to playlist
- Rename

and modal window:

- At the end of queue
- At the start of queue
- After the current song
- **Replace the queue**
- Replace the queue and play

some items are overlapping like _Replace queue_ and there is no _Replace the queue and play_ in the context menu

Would it to be better to combine all items in one single menu? 
I also want trigger such menu with right click and also shortcut.

mierak:
Assuming by "modal window" you mean the `AddOptions`

I am not going to put every single possibility of AddOptions into the context menu, it is there to provide some "basic" actions, not to solve every problem imaginable. You get rename, add, create playlist all in one menu whereas AddOptions "just" manipulates the queue.

> I also want trigger such menu with right click and also shortcut.

The context menu is triggered by both right click and a keybind. I noted this in the PR description...

Solarunit:
>Assuming by "modal window" you mean the AddOptions

yes


>I am not going to put every single possibility of AddOptions into the context menu, it is there to provide some "basic" actions, not to solve every problem imaginable. You get rename, add, create playlist all in one menu whereas AddOptions "just" manipulates the queue.


sane defaults, I get it. I wonder why `Add to queue` in the context menu for **playlists** when we have `Replace queue` already? I suggest to replace it with _Replace the queue and play_ instead.


>The context menu is triggered by both right click and a keybind. I noted this in the PR description...

`<C-z>` doesn't work here

mierak:
> doesn't work here

provide more info, kitty, ghostty, wezterm, foot, konsole all open context when right clicking on an item in queue/browsers for me

> <C-z> doesn't work here

put the bind in your config, it is merely a default in the generated config or when config is not found, by having an existing config you override all the keybinds

> sane defaults, I get it. I wonder why Add to queue in the context menu for playlists when we have Replace queue already? I suggest to replace it with Replace the queue and play instead.

two reasons, 1. the context menu is generated in the same way in one place for all the browser panes and 2. why not? I don't use replace queue basically ever, so if it was my way I would keep just add and remove the replace option, million different people, million different preferences

Solarunit:
>put the bind in your config, it is merely a default in the generated config or when config is not found, by having an existing config you override all the keybinds

I generated a new config and copied keybinds to my current config. Solved.

>two reasons, 1. the context menu is generated in the same way in one place for all the browser panes and 2. why not? I don't use replace queue basically ever, so if it was my way I would keep just add and remove the replace option, million different people, million different preferences


I see context menu is the same for Directories too. IMO it should be different.
For example `Add to playlist` in Playlists tab doesn't make much sense. And `Create playlist` just duplicates selected playlist instead of creating a new empty playlist.

Having `Add to queue` in the playlist tab is weird to me because you add some Playlist on top of existing queue (in my case on top of some loaded before playlist) :shrug:

mierak:
> I see context menu is the same for Directories too. IMO it should be different.
For example Add to playlist in Playlists tab doesn't make much sense. And Create playlist just duplicates selected playlist instead of creating a new empty playlist.

I don't see a problem with it duplicating a playlist, what if I want to do that? Now there is a way and before there was not. To be fair I really like this option because it is very useful to me so I am keeping it.

What should be different in the Directories exactly?

> Having Add to queue in the playlist tab is weird to me because you add some Playlist on top of existing queue (in my case on top of some loaded before playlist) 🤷

Again, that is weird to *you*. I mix and match playlists, I dont's see them as "separate queues".

What about the rightclick menu?

Solarunit:
>I don't see a problem with it duplicating a playlist, what if I want to do that? Now there is a way and before there was not. To be fair I really like this option because it is very useful to me so I am keeping it.

Maybe rename it to "Duplicate playlist"?
And there should be "Create new empty playlist" then.

>What should be different in the Directories exactly?

Context menu in Directories looks fine to me

Solarunit:
Also noticed, "Delete Playlist" option is missing

mierak:
> What about the rightclick menu?

I take it it works then?

> Maybe rename it to "Duplicate playlist"?
> And there should be "Create new empty playlist" then.

> Also noticed, "Delete Playlist" option is missing

https://github.com/mierak/rmpc/blob/6de833c2bbefad63c7515d7fbeb13d29b05110ab/src/ui/browser.rs#L575-L607

Feel free to submit a PR, I already dedicate several hours each day to rmpc on top of my 9-5 job and simply decided this was out of scope for this PR as it was already huge.

Solarunit:
I am not very familiar with rust :slightly_smiling_face: 

>I take it it works then?

rightclick menu works, two problems with it:

1. you have to double-click on the items
2. no selection on hover

mierak:
Honestly this is something that is going to be very subjective to each person. Problem with mouse integration in terminals is that there is no real "standard" way to do things like there is in GUI apps.

To give some exmaples:
- in yazi you cant even close or confirm the delete modal with mouse
- in lazygit, interacting with the help modal works the same way it does in rmpc now (except their scrolling is better)

I for one find the current state more intuitive, that is why it is like this in the first place and it is consistent throughout the program. If I were to change it to be your way I would get the next person complaining that it should be changed back.

Solarunit:
What do you think about tmux context menu?

- you can open it in any area of the screen next to the cursor
- to open it - just right click and hold
- when holding right click you can navigate through the menu and trigger items (no need to use left mouse button at all)
- to close menu just release right mouse button outside of the menu window

gif:

tmux-menus

mierak:
> you can open it in any area of the screen next to the cursor

I dont mind this, its just more work than reusing an existing modal

> to open it - just right click and hold

Having to hold right click is bad and unintuitive in my opinion

> when holding right click you can navigate through the menu and trigger items (no need to use left mouse button at all)

Same as above, an item being confirmed without explicit left click is bad, also how does keyboard interact with this? I dont mind items getting focused on hover if it is a proper context menu and not a reused modal, that is fine.

> to close menu just release right mouse button outside of the menu window

Again, it is bad and unintuitive, explicit confirmation of an item and of closing the menu by clicking either cancel or outside the menu is a must in my opinion.

All in all, please keep in mind the introduction sentence in the [mouse docs](https://mierak.github.io/rmpc/next/guides/mouse_controls/) - "These are meant to be complementary to the keybinds, not to replace keyboard controls". Rmpc is keyboard first, mouse support is just there for convenience, controlling everything via mouse is a non-goal for rmpc.

With that said, I will not stop you from implementing this if you want the menu to work closer to what tmux does with the points I have made above in mind. But expect this PR to be big.

Solarunit:
>Having to hold right click is bad and unintuitive in my opinion

It was strange at first but now I kinda like it.
But you have to be very careful when selecting items with this method.

>Same as above, an item being confirmed without explicit left click is bad, also how does keyboard interact with this?

If context menu was triggered with mouse, you can't interact with it, but if it was triggered with a shortcut I am not sure..

>Again, it is bad and unintuitive, explicit confirmation of an item and of closing the menu by clicking either cancel or outside the menu is a must in my opinion.

btw you can't close current menu by clicking outside of the menu

____
What do you think of changing items activation on single click instead of double click? Maybe as option?
(though I see it's not good without highlighting on hover)

PR #540, opened 2025-07-12
Feat/scrollbar mouse support

### LINKED ISSUE #475: Enable Mouse Interaction with Playlist Scrollbar
### Description

Currently, the scrollbar in the playlist view of rmpc is only visual—it shows the current position within the playlist, but it cannot be interacted with using the mouse. Attempting to click or drag the scrollbar does not scroll the playlist or change the view position.

Feature Request:
Please add support for mouse interaction with the playlist scrollbar. Ideally, users should be able to:
Click and drag the scrollbar to scroll through the playlist.
Click on a position within the scrollbar track to jump to that part of the playlist.

Rationale:
This would make navigation more intuitive, especially for users who prefer using the mouse.
The current visual-only scrollbar can be confusing, as it looks interactive but isn’t.
Many terminal-based UIs (and graphical ones) allow this, so it would improve usability and accessibility.

Additional Context:
I’m using rmpc in a terminal that supports mouse input.
The scrollbar’s position is helpful, but being able to interact with it would make large playlists much easier to browse.
Thank you for your work on rmpc! This feature would make the user experience even better.

### Proposed solution

Proposed Solution:
Allow users to interact with the playlist scrollbar using the mouse. For example, clicking and dragging the scrollbar should scroll through the playlist, and clicking on a spot in the scrollbar track should jump to that part of the playlist. This would make it easier to navigate long playlists without using only the keyboard.

Gri-ffin:
## Description

This PR implements mouse interaction with scrollbars in the browser panes (playlists, queue, etc.). 

Drag and Click support was also added in the mouse event system. Should not mess with keyboard navigation and existing mouse interactions and should work along with it.

**It should be noted that the scrollbar is only interactive when the content height exceeds the viewport. I made this intentional since if all items fit, there's nothing to scroll, and enabling interaction could lead to weird or confusing behavior. The scrollbar still shows position for consistency, but won’t respond to clicks in that case.**

### Related issues

fixes #475 

### Checklist

- [x] All tests passed


src/ui/browser.rs
```diff
@@ -157,7 +157,115 @@ where
...
+                    vertical: 0,
+                    horizontal: scrollbar_track.is_empty().into(),
+                }
+            }
+            Some(_) | None => ratatui::layout::Margin { vertical: 0, horizontal: 0 },
+        };
+
+        match event.kind {
+            MouseEventKind::LeftClick | MouseEventKind::Drag => {
+                if prev_area.width > 0 {
```
mierak, line 186: Previous stack dir should never be modified in this way. It will break things. I dont think the previous area should be scrollable at all.

src/ui/browser.rs
```diff
@@ -157,7 +157,115 @@ where
...
+                                #[allow(
+                                    clippy::cast_precision_loss,
+                                    clippy::cast_possible_truncation,
+                                    clippy::cast_sign_loss
+                                )]
+                                let target = (position_ratio
+                                    * (content_len.saturating_sub(1)) as f64)
+                                    .round() as usize;
+                                target.min(content_len.saturating_sub(1))
+                            };
```
mierak, line 246: Can we extract this logic to its own function in a way that would be usable everywhere? Say something like a function that takes scrollbars area, the drag event, content length and returns an `Option<usize>` where `Some` the event was handled and the index should be selected.

src/shared/mouse_event.rs
```diff
@@ -74,6 +77,7 @@ pub enum MouseEventKind {
     RightClick,
     ScrollDown,
     ScrollUp,
+    Drag,
```
mierak, line 80: I think we should track drag start position. This way we can then require drag events on the scrollbar to only require x coordinate to be the same at the start of the drag. Makes for a nicer and more natural user experience imho.

src/ui/browser.rs
```diff
@@ -157,7 +157,115 @@ where
...
+        let scrollbar_margin = match ctx.config.theme.scrollbar.as_ref() {
+            Some(scrollbar) if ctx.config.theme.draw_borders => {
+                let scrollbar_track = &scrollbar.symbols[0];
+                ratatui::layout::Margin {
+                    vertical: 0,
+                    horizontal: scrollbar_track.is_empty().into(),
+                }
+            }
+            Some(_) | None => ratatui::layout::Margin { vertical: 0, horizontal: 0 },
+        };
```
mierak, line 182: Seems like this is off by one with `draw_borders: false` and by two with `draw_borders: true`.

The scrollbars in the browser are quite a mess and I should redo them at some point ☹️

mierak (commented):
Thanks!

Aside from the few comments, I don't mind this not being done for the modals in this PR, we can always do that in the future. But I would like to see it implemented for all the major areas in the browser, meaning for the result phase of search pane as well.

> It should be noted that the scrollbar is only interactive when the content height exceeds the viewport

That is reasonable

src/ui/browser.rs
Gri-ffin: sure, i'll move it to `mouse_event.rs`

src/shared/mouse_event.rs
Gri-ffin: absolutely

src/ui/browser.rs
Gri-ffin: i guess then i can apply i temporary fix for it until redesign, leaving it buggy like that would bad, i can change it to be less smart like so:
```rust
let scrollbar_area = if ctx.config.theme.draw_borders {
    current_area.inner(ratatui::layout::Margin { vertical: 1, horizontal: 1 })
} else {
    current_area
};
```
this would still technically be a temporary fix and should be fixed during the redesign.
If you don't mind this temp fix i can push it.

mierak: I think the safer bet is to store the scrollbar's area directly when it gets rendered along with the current area in the areas array. Then we just need to check whether scrollbar is Some and its area.wdith > 0, then operate on the rightmost column of that area. Same goes for the search pane.

This way it will be safe from future refactors as well.

mierak: I think the safer bet is to store the scrollbar's area directly when it gets rendered along with the current area in the areas array. Then we just need to check whether scrollbar is Some and its area.wdith > 0, then operate on the rightmost column of that area. Same goes for the search pane.

This way it will be safe from future refactors as well.

src/shared/mouse_event.rs
```diff
@@ -67,6 +102,7 @@ pub struct MouseEvent {
     pub x: u16,
     pub y: u16,
     pub kind: MouseEventKind,
+    pub drag_start_position: Option<Position>,
```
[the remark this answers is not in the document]
Gri-ffin: i did it this just for simplicity, having it in the struct is more simple to use, and would reduce boilerplate, but i agree it would be more appropriate to have in the enum to be strict, since misuse of it in the code can produce weird executions in runtime which would not be catched by the tests.

mierak:
Its mostly there but now the scrollbar area is always `Rect::default()` no?

Im going to help a bit if you dont mind and drop a small patch. This should make all the scrollbars behave as I would expect. Essentially we expand the areas array in the browser struct directly, theres no real need to handle it separately for each pane so this mostly just removes that change.

One consideration to make here is to switch from indices to `EnumMap` like its done in the queue, but its fine like this.

<details>

<summary>diff</summary>

```diff
diff --git a/src/ui/panes/albums.rs b/src/ui/panes/albums.rs
index 5c55b6d..ba828a2 100644
--- a/src/ui/panes/albums.rs
+++ b/src/ui/panes/albums.rs
@@ -357,15 +357,6 @@ impl BrowserPane<DirOrSong> for AlbumsPane {
     }
 
     fn browser_areas(&self) -> [Rect; 4] {
-        let mut areas = [Rect::default(); 4];
-        let browser_areas = self.browser.areas;
-        for (i, area) in browser_areas.iter().enumerate().take(areas.len()) {
-            areas[i] = *area;
-        }
-        areas
-    }
-
-    fn scrollbar_area(&self) -> Option<Rect> {
-        self.browser.areas.get(2).copied()
+        self.browser.areas
     }
 }
diff --git a/src/ui/panes/directories.rs b/src/ui/panes/directories.rs
index 010eeab..09b83aa 100644
--- a/src/ui/panes/directories.rs
+++ b/src/ui/panes/directories.rs
@@ -410,15 +410,6 @@ impl BrowserPane<DirOrSong> for DirectoriesPane {
     }
 
     fn browser_areas(&self) -> [Rect; 4] {
-        let mut areas = [Rect::default(); 4];
-        let browser_areas = self.browser.areas;
-        for (i, area) in browser_areas.iter().enumerate().take(areas.len()) {
-            areas[i] = *area;
-        }
-        areas
-    }
-
-    fn scrollbar_area(&self) -> Option<Rect> {
-        self.browser.areas.get(2).copied()
+        self.browser.areas
     }
 }
diff --git a/src/ui/panes/playlists.rs b/src/ui/panes/playlists.rs
index e73d3da..34d2719 100644
--- a/src/ui/panes/playlists.rs
+++ b/src/ui/panes/playlists.rs
@@ -598,15 +598,6 @@ impl BrowserPane<DirOrSong> for PlaylistsPane {
     }
 
     fn browser_areas(&self) -> [Rect; 4] {
-        let mut areas = [Rect::default(); 4];
-        let browser_areas = self.browser.areas;
-        for (i, area) in browser_areas.iter().enumerate().take(areas.len()) {
-            areas[i] = *area;
-        }
-        areas
-    }
-
-    fn scrollbar_area(&self) -> Option<Rect> {
-        self.browser.areas.get(2).copied()
+        self.browser.areas
     }
 }
diff --git a/src/ui/panes/search.rs b/src/ui/panes/search.rs
index bb2666d..e8f1edd 100644
--- a/src/ui/panes/search.rs
+++ b/src/ui/panes/search.rs
@@ -54,7 +54,7 @@ pub struct SearchPane {
     preview: Option<Vec<PreviewGroup>>,
     songs_dir: Dir<Song>,
     input_areas: Rc<[Rect]>,
-    column_areas: [Rect; 3],
+    column_areas: [Rect; 4],
 }
 
 const PREVIEW: &str = "preview";
@@ -84,7 +84,7 @@ impl SearchPane {
                 [ButtonInput { label: " Reset", variant: ButtonInputVariant::Reset }],
             ),
             input_areas: Rc::default(),
-            column_areas: [Rect::default(); 3],
+            column_areas: [Rect::default(); 4],
         }
     }
 
@@ -160,10 +160,16 @@ impl SearchPane {
         let inner_block = block.inner(area);
 
         self.column_areas[1] = inner_block;
+        self.column_areas[3] =
+            if matches!(self.phase, Phase::BrowseResults { .. }) { area } else { Rect::default() };
         frame.render_widget(block, area);
         frame.render_stateful_widget(current, inner_block, directory.state.as_render_state_ref());
         if let Some(scrollbar) = config.as_styled_scrollbar() {
-            frame.render_stateful_widget(scrollbar, area, directory.state.as_scrollbar_state_ref());
+            frame.render_stateful_widget(
+                scrollbar,
+                self.column_areas[3],
+                directory.state.as_scrollbar_state_ref(),
+            );
         }
     }
 
@@ -917,7 +923,7 @@ impl SearchPane {
     }
 
     fn scrollbar_area(&self) -> Option<Rect> {
-        let area = self.column_areas[1];
+        let area = self.column_areas[3];
         if area.width > 0 { Some(area) } else { None }
     }
 
diff --git a/src/ui/panes/tag_browser.rs b/src/ui/panes/tag_browser.rs
index 07ea927..6d2131b 100644
--- a/src/ui/panes/tag_browser.rs
+++ b/src/ui/panes/tag_browser.rs
@@ -649,8 +649,7 @@ impl BrowserPane<DirOrSong> for TagBrowserPane {
     }
 
     fn browser_areas(&self) -> [Rect; 4] {
-        let [a, b, c] = self.browser.areas;
-        [a, b, c, Rect::default()]
+        self.browser.areas
     }
 }
 
diff --git a/src/ui/widgets/browser.rs b/src/ui/widgets/browser.rs
index 4b3f599..fb3d63d 100644
--- a/src/ui/widgets/browser.rs
+++ b/src/ui/widgets/browser.rs
@@ -12,7 +12,7 @@ use crate::{
 #[derive(Debug)]
 pub struct Browser<T: std::fmt::Debug + DirStackItem + Clone + Send> {
     state_type_marker: std::marker::PhantomData<T>,
-    pub areas: [Rect; 3],
+    pub areas: [Rect; 4],
     filter_input_active: bool,
 }
 
@@ -20,7 +20,7 @@ impl<T: std::fmt::Debug + DirStackItem + Clone + Send> Browser<T> {
     pub fn new() -> Self {
         Self {
             state_type_marker: std::marker::PhantomData,
-            areas: [Rect::default(); 3],
+            areas: [Rect::default(); 4],
             filter_input_active: false,
         }
     }
@@ -168,11 +168,13 @@ where
                 state.as_render_state_ref(),
             );
             self.areas[1] = inner_block;
+            let scrollbar_area = current_area.inner(scrollbar_margin);
+            self.areas[3] = scrollbar_area;
             ratatui::widgets::Widget::render(block, current_area, buf);
             if let Some(scrollbar) = config.as_styled_scrollbar() {
                 ratatui::widgets::StatefulWidget::render(
                     scrollbar,
-                    current_area.inner(scrollbar_margin),
+                    scrollbar_area,
                     buf,
                     state.as_scrollbar_state_ref(),
                 );
```

</details>

Gri-ffin:
I don't mind at all, thanks for the help.

Gri-ffin:
should be good now

mierak:
looks good, thanks!

PR #551, opened 2025-07-23
docs: Add step for configuring Cava pane in layout

PolvosMagicos:
## Description

Just an extra step on the docs to show cava after configuring it on the pane layout

### Checklist

- [ ] All tests passed


mierak:
Right, I forgot about that. Thanks!

PR #563, opened 2025-07-31
Support adding marked songs to a playlist

ppebb:
## Description
Adds support for adding multiple songs to playlists from the queue. I referenced implementations for adding one song and deleting multiple, but I'm largely unfamiliar with Rust so I may have made some mistakes. Appears functional from my own testing, though.

Also unsure if I need to update documentation. The binding for `<Space>` is described as "useful for adding multiple songs to a playlist," (although that seemingly did not work from the queue prior to writing this PR) so I believe this functionality is already documented; However, you may want to modify the description for the queue binding for `a`.

### Related issues
None that I know of.

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ going to the next item
 - Scrollbars now represent the viewport position instead of the currently selected item position
 - Remote commands now check for `$PID` env variable, meaning `--pid` argument is no longer needed for
 remote commands inside scripts triggered by rmpc
+- `AddToPlaylist` binding handles marked songs rather than only the one under your cursor.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


mierak:
Thanks!

I planned to implement this, though a bit differently as a part of these two issues: https://github.com/mierak/rmpc/issues/531 https://github.com/mierak/rmpc/issues/493. Though I see no harm in merging this in the meantime as a stopgap measure.

No issues with the code, just need to appease clippy so the build passes.

ppebb:
Swear I checked clippy locally, guess I missed something. Will have it fixed shortly.

mierak:
thanks!

Issue #564, opened 2025-08-01
Dynamic colors


osamaaj03:
### Description

Hey! You're doing an amazing job! Sorry for being a bit quiet lately; I didn't have much to add since 0.9 is almost feature complete for me. Anyway, I'll cut to the chase—what I'm asking for is something like this:

https://github.com/user-attachments/assets/1a74237f-9360-4879-9901-9d12d65d0a1b

### Proposed solution

Someone smarter than me could probably figure out how [wallust](https://codeberg.org/explosion-mental/wallust) does it and implement something similar

mierak:
Hi!

If I understand you correctly, you want to synthesize a color scheme from the album art and apply it to rmpc I assume?

That would be quite an undertaking, isn't exactly this already possible using wallust which you referenced? Use `on_song_change` to call `rmpc albumart` and save it to some temp dir, then utilize wallust to use that image as an input to synthesize rmpc's theme from a template? Rmpc would then autoreload the theme. (Or maybe that is exactly what you've done in the video?)

A native implementation would be possible, but it would be basically the same mechanism utilizing the current config hot reloading except happening only internally.

osamaaj03:
>Or maybe that is exactly what you've done in the video?

Almost! I’m using pywal instead of wallust just because it has more templates, plus a wrapper for rmpc to handle the on_enter and on_exit commands. It also has a script to change the colors of the visualizer and notifications, which is quite a lot. That’s one reason I want this to be implemented natively.

> but it would be basically the same mechanism utilizing the current config hot reloading except happening only internally.

A smoother transition would be awesome! The way the colors shift in my setup is pretty jittery and asynchronous, and it can get a bit buggy sometimes.

I'm sure users would love having this functionality as a simple toggle with a boolean instead of dealing with all this:

```bash
#!/bin/bash

# Song information
ALBUMARTIST="$(jq -r .metadata.albumartist <(rmpc song))"
TITLE="$(jq -r .metadata.title <(rmpc song))"

# Directories where to store temporary data
TMP_DIR="/tmp/rmpc"

# Ensure the directories is created
mkdir -p "$TMP_DIR"

# Where to temporarily store the album art received from rmpc
ALBUM_ART_PATH="$TMP_DIR/notification_cover"

# Where to temporarily store the album art colors
ALBUM_COLORS_PATH="$TMP_DIR/wal"

# Save album art of the currently playing song to a file
rmpc albumart --output "$ALBUM_ART_PATH"

# Change the colors
[...]

PR #566, opened 2025-08-02
feat: run `on_song_change` at startup

osamaaj03:
## Description

Like ncmpcpp does it! 

### Related issues

#564

### Checklist

- [x] All tests passed

### DIFF
--- a/src/core/event_loop.rs
+++ b/src/core/event_loop.rs
@@ -75,6 +75,19 @@ fn main_task<B: Backend + std::io::Write>(
         }
     };
 
+    // Check if there is a current song playing on startup
+    if let Some((_, song)) = ctx.find_current_song_in_queue() {
+        if let Some(command) = &ctx.config.on_song_change {
+            let mut env = create_env(&ctx, std::iter::empty());
+
+            if let Some(played) = ctx.song_played {
+                env.push(("PREV_ELAPSED".to_owned(), played.as_secs().to_string()));
+            }
+
+            run_external(command.clone(), env);
+        }
+    }
+
     match ctx.status.state {
         State::Play => {
             // Start update loop since a song is playing on startup
--- a/src/core/event_loop.rs
+++ b/src/core/event_loop.rs
@@ -76,14 +76,10 @@ fn main_task<B: Backend + std::io::Write>(
     };
 
     // Check if there is a current song playing on startup
-    if let Some((_, song)) = ctx.find_current_song_in_queue() {
-        if let Some(command) = &ctx.config.on_song_change {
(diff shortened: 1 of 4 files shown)


src/core/event_loop.rs
```diff
@@ -75,6 +75,19 @@ fn main_task<B: Backend + std::io::Write>(
...
     };
 
+    // Check if there is a current song playing on startup
+    if let Some((_, song)) = ctx.find_current_song_in_queue() {
+        if let Some(command) = &ctx.config.on_song_change {
+            let mut env = create_env(&ctx, std::iter::empty());
+
+            if let Some(played) = ctx.song_played {
+                env.push(("PREV_ELAPSED".to_owned(), played.as_secs().to_string()));
+            }
```
mierak, line 85: This does not make sense on startup and is useless without `$PREV_SONG` as well.

src/core/event_loop.rs
```diff
@@ -75,6 +75,19 @@ fn main_task<B: Backend + std::io::Write>(
         }
     };
 
+    // Check if there is a current song playing on startup
+    if let Some((_, song)) = ctx.find_current_song_in_queue() {
+        if let Some(command) = &ctx.config.on_song_change {
```
mierak, line 80: Just a cosmetic nitpick since if let chains are finally stable in rust. Collapse the nested if let to a single block to reduce nesting.

```suggestion
    if let Some((_, song)) = ctx.find_current_song_in_queue()
        && let Some(command) = &ctx.config.on_song_change {
```

mierak (commented):
Thanks! Aside from the two small comments, can we have this behind a option which defaults to false? Something as simple as

```rust
#[serde(default)]
run_on_song_change_on_start: bool, // even better if you can think of a better name
```

I see the merit of this, but myself prefer it to not execute on startup.

osamaaj03:
>even better if you can think of a better name

`kickstart_my_on_song_change` could be fun

>can we have this behind an option which defaults to false?

I don't see any problem with that! I'm just away from my laptop at the moment, but I'll check it as soon as I'm back home.

mierak:
> kickstart_my_on_song_change could be fun

I prefer something more descriptive, the one I suggested is fine, not great but fine. "kickstart" leaves a lot to interpretation.

osamaaj03:
It was worth a shot! yeah fyi I didn't test or format the new commit yet.

osamaaj03:
I went with `exec_on_song_change_at_start`—what do you think?

mierak:
> exec_on_song_change_at_start

I like it, way more than what I came up with, thanks!

Looks good to me aside from clippy's yapping.

mierak:
thanks!

Issue #568, opened 2025-08-02
Configurable date tag when browsing by album artist / artist / genre


jebaum:
### Description

The rmpc panes for album artist, artist and genre use the `date` tag in various spots both for display, and for sorting albums. For a lot of albums, the `date` tag won't have the actual date the album was released, generally for remasters that came out in later years. The `originaldate` tag may exist and be filled in with the original album release date. This is what I'd prefer to be displayed personally.

### Proposed solution

The tag used could be made configurable, so users can choose whether they want to use the `date` tag, `originaldate`, or something else. The configuration could take a list of tags, falling back to later entries in the list if earlier entries aren't present in the album.

I have a patch for this locally which is working for me. It's the first rust I've ever written so it may be garbage, but happy to share it in a PR as a starting point.

Here's a screenshot showing rmpc running with my patch on the upper screen, and without it on the lower screen:

<img width="2159" height="1917" alt="Image" src="https://github.com/user-attachments/assets/b5955759-d054-4bf8-b675-652bfdcaae83" />

mierak:
This could be simply added into the [artists](https://mierak.github.io/rmpc/next/configuration/#artists) block in the config (the fact that the artists block should be renamed aside, thats a separate issue...). Making it take a list of tags with fallbacks like you suggest also sounds fine, though what values should be allowed in there? I think only the two (date, originaldate) make sense right?

> I have a patch for this locally which is working for me. It's the first rust I've ever written so it may be garbage, but happy to share it in a PR as a starting point.

I am always happy to take contributions and dont mind chiming in if the code itself needs some work!

PS: lovely theme, finally something more original!

jebaum:
Yeah those are the only two I know of. So probably the PR I opened can either be made simpler, or something has to be done to check that there isn't a disallowed value present

PR #569, opened 2025-08-02
support originaldate tag when browsing

jebaum:
## Description
Allow album artists + artist + genre panes to support configurable date tags with fallback behavior.

In my `config.ron` to use this I have:

```
    artists: (
        album_display_mode: SplitByDate,
        album_sort_by: Date,
        album_date_tags: [OriginalDate, Date], // changed during PR to not be strings
    ),
```

### Related issues
https://github.com/mierak/rmpc/issues/568

### Checklist

- [x] All tests passed

### DIFF
--- a/src/config/artists.rs
+++ b/src/config/artists.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 pub struct Artists {
     pub album_display_mode: AlbumDisplayMode,
     pub album_sort_by: AlbumSortMode,
-    pub album_date_tags: Vec<String>,
+    pub album_date_tags: Vec<AlbumDateTag>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -13,8 +13,8 @@ pub struct ArtistsFile {
     pub album_display_mode: AlbumDisplayMode,
     #[serde(default)]
     pub album_sort_by: AlbumSortMode,
-    #[serde(default = "default_album_date_tags")]
-    pub album_date_tags: Vec<String>,
+    #[serde(default)]
+    pub album_date_tags: Vec<AlbumDateTag>,
 }
 
 impl Default for ArtistsFile {
@@ -22,7 +22,7 @@ impl Default for ArtistsFile {
         Self {
             album_display_mode: AlbumDisplayMode::default(),
             album_sort_by: AlbumSortMode::default(),
-            album_date_tags: default_album_date_tags(),
+            album_date_tags: vec![AlbumDateTag::Date],
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/config/artists.rs
```diff
@@ -1,17 +1,30 @@
...
 
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ArtistsFile {
     #[serde(default)]
     pub album_display_mode: AlbumDisplayMode,
     #[serde(default)]
     pub album_sort_by: AlbumSortMode,
+    #[serde(default = "default_album_date_tags")]
+    pub album_date_tags: Vec<String>,
```
mierak, line 17: I think this is better represented as an enum
```suggestion
    pub album_date_tags: Vec<AlbumDateTag>,
```

and then impl the enum and adjust the default etc accordingly: 

```rust
#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, strum::IntoStaticStr)]
#[strum(serialize_all = "lowercase")]
pub enum AlbumDateTag {
    #[default]
    Date,
    OriginalDate,
}
```

This will ensure that no other value can get in there via config file or other means, the `strum::IntoStaticStr` coupled with the `lowercase` attribute will enable you to get the enum value as a `&'static str` - to be used when getting values out of the map.

src/config/artists.rs
```diff
@@ -28,8 +41,26 @@ pub enum AlbumSortMode {
...
+    vec!["date".to_string()]
+}
+
+impl Default for Artists {
+    fn default() -> Self {
+        Self {
+            album_display_mode: AlbumDisplayMode::default(),
+            album_sort_by: AlbumSortMode::default(),
+            album_date_tags: default_album_date_tags(),
+        }
```
mierak, line 54:
```suggestion
        ArtistsFile::default().into()
```

Better to simply proxy the other struct's default impl to ensure that they have same defaults.

src/ui/panes/tag_browser.rs
```diff
@@ -199,10 +199,16 @@ impl TagBrowserPane {
                 let album = song.metadata.get("album").map_or("<no album>".to_string(), |v| {
                     v.join(&ctx.config.theme.format_tag_separator).to_string()
                 });
-                let song_date = song.metadata.get("date").map_or("<no date>", |v| v.last());
+                let song_date = ctx
+                    .config
+                    .artists
+                    .album_date_tags
+                    .iter()
+                    .find_map(|tag| song.metadata.get(tag).map(|v| v.last().to_string()))
```
mierak, line 207: Here we can then use the derived `IntoStaticStr` from the enum.

```suggestion
                    .find_map(|tag| {
                        song.metadata
                            .get(Into::<&'static str>::into(tag))
                            .map(|v| v.last().to_string())
                    })

```

mierak (commented):
The code itself is completely fine, all my comments are simply about changing String to a more constrained enum in the config.

jebaum:
Think I covered it all. Thanks for walking me through it!

src/config/artists.rs
```diff
@@ -1,17 +1,30 @@
...
 }
 
-#[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ArtistsFile {
     #[serde(default)]
     pub album_display_mode: AlbumDisplayMode,
     #[serde(default)]
     pub album_sort_by: AlbumSortMode,
+    #[serde(default)]
```
mierak, line 16: this needs to keep the custom default function though unfortunatelly, because default for `Vec` is empty Vec, meaning it would change the behavior for existing configs to not use any date tag
```suggestion
    #[serde(default = "default_album_date_tags")]
```

mierak:
Other than this last thing it looks ok, thanks!

PR #578, opened 2025-08-11
Add addyt --name to search by song title

PolvosMagicos:
## Description
- Introduces --name to addyt to resolve the first YouTube match and enqueue via existing download pipeline.
- Works in CLI and TUI Command Mode (custom argv splitter, no extra deps).

### Checklist

- [x] All tests passed

### DIFF
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -201,12 +201,24 @@ pub enum Command {
     },
     /// Add a song from youtube to the current queue.
     AddYt {
-        url: String,
+        #[arg(
+            required_unless_present = "name",
+            value_parser = clap::builder::NonEmptyStringValueParser::new()
+        )]
+        url: Option<String>,
         /// If provided, queue the new item at this position instead of the end
         /// of the queue. Allowed positions are <number> (absolute) and
         /// +<number> or -<number> (relative)
         #[arg(short, long, allow_negative_numbers = true)]
         position: Option<QueuePosition>,
+        /// If provided search by name and use the first result
+        #[arg(
+            short = 'n',
+            long = "name",
+            conflicts_with = "url",
+            value_parser = clap::builder::NonEmptyStringValueParser::new()
+        )]
+        name: Option<String>,
     },
     /// List MPD outputs
     Outputs,
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -444,4 +444,87 @@ impl Args {
         path.push("config.ron");
         path
     }
+
+    /// Split a shell-like command line into tokens (argv).
+    ///
+    /// Supports single `'...'` and double `"..."` quotes and backslash escapes.
+    ///
+    /// # Errors
+    /// Returns `Err` if the input contains an **unclosed quote** or a **dangling backslash**
+    /// escape at the end of the string.
+    ///
+    /// # Examples
+    /// ```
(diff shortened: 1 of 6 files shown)


mierak:
Thanks, apologies for not getting to this sooner. I have two questions:
1. Why is the custom args parser needed exactly? I assume because the current implementation splits naively at whitespace? The current implementation should be removed if that is the case then.
2. `addyt` was expanded over time to support multiple streaming hosts (yt, soundcloud and niconico currently). I think its relatively confusing to have --name param to only work for one of the hosts and dont see a way to make this consistent in some way. Wouldn't it be better to have a separate command for this? Say findyt or searchyt which could support these hosts in the future as well with an arg (which could be made optional and default to youtube).

And just some bonus thinking: I imagine this could also be nice and relatively simple to expand in the future in a way that would show you first N entries and let you choose which one to add/download. Maybe --first/-f arg or something of that sort should be forced in the current implementation to allow for this when the arg is not provided?

PolvosMagicos:
Thanks for the review, about those 3 points

1. Yes, Command Mode previously split on whitespace, so quoted args (e.g. --name "cool song") broke. The custom splitter is scoped to Command Mode only; the normal CLI is unchanged. I kept it local to minimize risk of breaking other modules.
2. Agreed it’s confusing. I can add a separate search/find command (yt) and keep addyt for direct URLs; deprecate --name.
3. Something like the nekotina bot on discord to choose from a list would be nice, maybe I can add that on a future PR.

mierak:
Yeah, I think it would be better to create a separate command, searchyt is fine I think. The current args parser is simple FromStr on Args, its only used in one place so it should be safe to delete. No need to deprecate anything since the --name was added in this PR.

src/config/cli.rs
```diff
@@ -208,6 +208,13 @@ pub enum Command {
         #[arg(short, long, allow_negative_numbers = true)]
         position: Option<QueuePosition>,
     },
+    SearchYt {
```
mierak, line 211: Some short description of what this command does would be nice

src/config/cli.rs
```diff
@@ -208,6 +208,13 @@ pub enum Command {
...
+        /// Search query (song/artist/title…)
+        #[arg(value_name = "QUERY")]
+        query: String,
+        #[arg(short, long, allow_negative_numbers = true)]
+        position: Option<QueuePosition>,
```
mierak, line 216: doc comment missing here as well, feel free to copy/edit docs from the existing position arguments on other commands

src/core/command.rs
```diff
@@ -268,6 +268,19 @@ impl Command {
...
+                let file_paths = YtDlp::init_and_download(config, &chosen_url)?;
+                Ok(Box::new(move |client| {
+                    client.send_start_cmd_list()?;
+                    for file in &file_paths {
+                        client.send_add(file, position)?;
+                    }
+                    client.send_execute_cmd_list()?;
+                    client.read_ok()?;
+                    Ok(())
+                }))
```
mierak, line 282: No issues with the code but actually using this feels a bit awkard since the program just hangs for some time and then exits successfully without printing anything. I think adding a few `eprintln` statements here could help, something like
```
searching yt
found ..., downloading...
```

Careful to not call print inside the returned closure though.

mierak (commented):
Just a few cosmetic/docs comments, otherwise it looks good to go after they are addressed, thanks!

PR #586, opened 2025-08-23
feat: Added an ability to specify relative/absolute theme path in config

Illuxel:
## Description

Hi! I made a small feature pr for allowing user to specify a relative path to a theme.ron file.

### Related issues

Can't specify absolute/relative theme path in config file with file extension. Only a theme name with hard coded directory 'themes' are allowed.

### Check list
- [x] All tests passed
- [x] Support for relative paths
- [x] Support for absolute paths
- [x] Support for legacy theme paths
- [x] Update changelog
- [x] Update docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
 - Config option for global lyrics offset - `lyrics_offset_ms`
 - add `exec_on_song_change_at_start` config option
 - Added `album_date_tags` config option to specify priority order of metadata tags for album dates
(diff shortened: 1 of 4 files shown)


mierak:
Hi!

A few issues with this PR.
* The change breaks existing configs.
* If this change is getting in I would like to see support for absolute paths as well and it should check for the existing behavior first.
* Documentation is missing
* Changelog is missing

illuxel:
Thanks for reply.

I added support for absolute paths and restored old behavior.

mierak:
Code needs to be formatted, after that its good to go.

PR #589, opened 2025-08-26
Add link-time optimization directives to Cargo.toml, for smaller binary size

Babkock:
## Description

Hi, this is my first time making a pull request here.  I really enjoy this music player, I want to help out with the issues I find with it.  I thought you might be interested in this, I don't know if you know about these.  [For reference.](https://doc.rust-lang.org/cargo/reference/profiles.html)

I added a few options to the `Cargo.toml` file, under the heading **`[profile.release]`** to allow **rmpc** to be built with a smaller binary size. These options are `strip = "symbols"` , `lto = "fat"` , and `opt-level = "s"`.

I did not mark "Changelog has been updated", as my change did not affect the behavior of the **rmpc** program itself. I do not feel it is necessary to add such a small change to the Changelog.

See the below screenshot.  The first command shows the file size of the **rmpc** binary on my system (`/usr/bin`), which was built with the current `Cargo.toml` on `master`.  The second command shows the file size of the binary built with my new **`[profile.release]`** optimizations.  The program works exactly the same way.

binarysize

### Checklist

- [X] All tests passed
- [X] Code has been formatted with nightly
- [X] Code has been checked with clippy (stable)
- [X] Documentation has been updated (if needed)

Thank you for your time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,4 +97,5 @@ unused_macros = "allow"
 [profile.release]
 strip = "symbols"
 lto = "fat"
+opt-level = "s"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,3 +93,8 @@ cast_sign_loss = "allow"
 
 [lints.rust]
 unused_macros = "allow"
+
+[profile.release]
+strip = "symbols"
+lto = "fat"
+


mierak:
Hi, Im glad you are enjoying rmpc! 

I am aware of LTO and other optimizations, however I believe that saving a few (3 in this case) MBs of storage for application like rmpc is not worth the additional compile time penalty this incurs. 

Some rough quick testing on my machine (16c/32t - 9950x3d, using mold), doing `cargo clean` followed by `cargo build --release` 
* current master - 11s average
* with these optimizations - 43s average

The quadrupling of the release compile time makes it hard to justify the binary size savings.

PR #628, opened 2025-09-14
searchyt: SoundCloud support + interactive picker (CLI & TUI)

PolvosMagicos:
## Description
This PR expands the searchyt flow in two ways:
1. adds SoundCloud search support.
3. introduces an interactive list picker for both terminal (CLI) and TUI so users can choose among the top N results.

### Checklist

- [x] All tests passed


src/ui/mod.rs
```diff
@@ -536,6 +570,52 @@ impl<'ui> Ui<'ui> {
...
+        query: &str,
+        soundcloud: bool,
+        limit: usize,
+        interactive: bool,
+        position: Option<QueuePosition>,
+    ) -> anyhow::Result<Option<Command>> {
+        let kind = if soundcloud { YtDlpHostKind::Soundcloud } else { YtDlpHostKind::Youtube };
+
+        if interactive {
+            let items = YtDlp::search_many(kind, query.trim(), limit)?;
```
mierak, line 584: One more thing, it would be great if this and the other search at L615 were also routed through worker thread because the TUI can freeze up if youtube is unresponsive for whatever reason which is undesirable.
But I can refactor that later if dont feel like threading the whole thing through.

PolvosMagicos:
Thanks for pointing those out, about the worker thread for the search, I notice it freezes up a bit, I'll take a look on the thread workers for that as well.

ghost:
WOAH that's exactly what i wanted to do. AWESOME

ghost:
Would it be possible to pass down more arguments?

I'd really like to have split-chapters (with exact cuts).

Playlist support would be really nice too, to only search for playlists or videos with timestamps

mierak:
> Would it be possible to pass down more arguments?
> 
> I'd really like to have split-chapters (with exact cuts).
> 
> Playlist support would be really nice too, to only search for playlists or videos with timestamps

I don't think these are in scope for this PR and even then, rmpc is MPD client first and foremost, not a yt-dlp frontend and we have to draw the line somewhere (yt-dlp has an astounding amount of options, its not feasible to implement every single one). These add/searchyt serve mostly as a convenience. I think its better to do work like this outside of rmpc and import your music.

ghost:
> > Would it be possible to pass down more arguments?
> > I'd really like to have split-chapters (with exact cuts).
> > Playlist support would be really nice too, to only search for playlists or videos with timestamps
> 
> I don't think these are in scope for this PR and even then, rmpc is MPD client first and foremost, not a yt-dlp frontend and we have to draw the line somewhere (yt-dlp has an astounding amount of options, its not feasible to implement every single one). These add/searchyt serve mostly as a convenience. I think its better to do work like this outside of rmpc and import your music.

alr I see, i thought it was quite simple to do. But i can see how it'd be out of scope as it'd have to change the yt-dlp functionallity for the whole rmpc client. It'd be neat as I always add each individual songs and its hard to know individual song names of an album.

I think searching for playlist would be specific to this PR though.

Also because you were adding providers would it be possible to add bandacamp it also works with ytdlp?

And it'd be nice to see the channel/uploader name when selecting the song.

mierak (commented):
Looks ok to me except this one small thign.

Issue #484, opened 2025-06-21
Sticker-based song rating ability


CyberShadow:
### Description

The unmaintained Cantata mpd client had a "rating" feature - a song's rating could be given a rating between 0.5 and 5 stars. This was persisted in mpd's sticker database (see e.g. `mpc` manual page, section "Stickers"), with the key `rating` and a numeric value between 1 and 10.

### Proposed solution

Would it be possible to have an option (or a plugin?) to enable similar functionality? I think at the least it would be nice to be able to view and set a song's rating.

mierak:
As a stopgap measure, rmpc already supports manipulating stickers via CLI and you can bind these commands to keybinds to for example do `ExternalCommand(command: ["sh", "-c", "rmpc sticker set \"$FILE\" rating 3"], description: "")` (untested, and you would also need a separate bind per rating level so it kind of sucks).
These stickers can then be displayed in the queue and property panes/header via `Sticker("rating")` property - [docs](https://mierak.github.io/rmpc/next/configuration/header/#sticker) and another [docs](https://mierak.github.io/rmpc/next/configuration/song-table/#sticker)
There also is an example of using stickers to [track playcount](https://mierak.github.io/rmpc/next/guides/on_song_change/#track-song-play-count)
```bash
$ rmpc sticker --help
Manipulate and query song stickers

Usage: rmpc sticker <COMMAND>

Commands:
  set        Set sticker value for a song
  get        Get sticker value for a song
  list       List all stickers of a song
  find       Find all stickers of given name in  the specified directory
  delete     Delete a sticker from a song
  deleteall  Delete all stickers in a song
  help       Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

What's missing is some first-class support for the rating where I am not sure about any specifics at the moment.

CyberShadow:
Thanks!

> What's missing is some first-class support for the rating where I am not sure about any specifics at the moment.

Happy to provide details for how it's done in Cantata. If it helps:

- Rating is measured in stars, with a granularity of 0.5 stars.
- The best possible rating is 5 stars (translating to a sticker value of `10`).
- The worst possible rating is 0.5 stars (translating to a sticker value of `1`).
- In Cantata, attempting to set the rating to 0 stars clears the rating instead, which suggests that a sticker value of `0` is synonymous with an absent `rating` sticker.

Gremious:
[...]

PR #633, opened 2025-09-15
feat: rating system

mierak:
## Description

- Adds a new keybind action: `Rate()`. Without any options set it will open a modal window which allows you to set rating on given song(s). The keybind can be further customized to set a rating directly or to customize what values are displayed in the modal - docs are missing on this still and the keybind will most likely still be expanded a little bit.
- Search pane now has rating filters (any/equal/greater than/less than) a given value. This is only visible when you have MPD configured to use stickers. You can search with any combination of filters and rating or just by rating.
- Stickers can now be displayed in the browser panes, meaning you can display song's rating in there. This was possible in the Queue pane till now.

A bunch of things happening here, some refactors to how previews are fetched and rendered. Stickers were moved to a separate map in ctx instead of in each song because their lifecycle is not inherently tied to the song. New modal section etc etc.

~~Very work in progress right now but~~ should set foundation for further stuff with stickers and further refinement of the dir stack which is kind of a mess right now. Also a fixes bunch of issues like previews sometimes not respecting browser song format and others.

### Related issues
partially addresses https://github.com/mierak/rmpc/issues/484

### Checklist

- [ ] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file.
 - Added `keep_state_on_song_change` and equivalent flag to cli
 - Added `ignore_leading_the` when sorting entries in browsers
 - More information about the system to debuginfo
+- Stickers are now usable in `browser_song_format`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ All notable changes to this project will be documented in this file.
 - Added `ignore_leading_the` when sorting entries in browsers
 - More information about the system to debuginfo
 - Stickers are now usable in `browser_song_format`
+- Rating support, rating can be set on a song. The rating can be displayed in the queue table and
+browser panes as well as searched for in the Search pane
 
 ### Changed
 
--- a/docs/src/content/docs/next/configuration/keybinds.mdx
+++ b/docs/src/content/docs/next/configuration/keybinds.mdx
@@ -152,12 +152,13 @@ list and controlling search mode.
 #### Rate
 
 A configurable keybind used to rate songs. Allows you to rate either the currently playing song or
-the song(s) under cursor in the browser panes.
+the song(s) under cursor in the browser panes and the queue.
 
 A modal window with options to select is displayed instead when no configuration is provided or if
 the keybind is directly configured to display a modal.
 
-Please check <a href={path("configuration/stickers#rating")}>stickers</a> page for more information.
+Must have stickers configured, please check <a href={path("configuration/stickers#rating")}>stickers</a>
+page for more information.
 
 :::note
 It is highly advised to keep ratings in the 0-10 range. Other MPD clients or some other rmpc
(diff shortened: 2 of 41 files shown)


Gremious:
> It is always a big help to get another set of eyes on stuff 

Two notes:

1. Bless, there's a "Clear rating" button so there's a clear distinction between 0 and Unrated / being able to remove accidental ratings. Thank you :)

2. Ratings don't seem to update in the UI immediately when applied, is that intentional?

I made a keybind like so: `"R": Rating(kind: Value(5))`
And I displayed it in a property pane:
```ron
(
size: "60%",
    pane: Pane(Property(
        content: [
            (kind: Property(Song(Title))),
            (kind: Text(" Rating: ")),
            (kind: Sticker("rating")),
        ], align: Right,
    )),
),

```

However, after pressing the keybind, the property text did not update. I moved around, played a different song, re-played this one, etc. - nothing.

I then went to the "Search" tab and scrolled through the rating section to force it to search for "Rating Lesss than 100" and that did trigger a "refresh", after which everything displays correctly.

I am not super duper familiar with rmpc yet, so, sorry if I'm missing something.

mierak:
Thanks a bunch for looking at this!

> Ratings don't seem to update in the UI immediately when applied, is that intentional?

No its not, its possible I missed a place. Can you tell me exactly what actions you took on which pane to reproduce this? It should always trigger a refresh no matter what.

> Bless, there's a "Clear rating" button so there's a clear distinction between 0 and Unrated / being able to remove accidental ratings. Thank you :)

Yep! It also is specifically useful if you want to see what songs you might not have rated yet etc.

Also I just now realized that making the action be able to rate the currently playing song instead of just the one under the cursor might pretty good too.

Gremious:
Here's a quick recording:

https://asciinema.org/a/sBzEWKGIiJOZlFbDfOVyGbQqW

Written step-by-step

1. Clone and checkout this branch
2. create rmpc config in current folder via `rmpc config > config.ron`, edit it to have the code mentioned above:

`keybinds: (navigation: { "R":         Rating(kind: Value(5)),})`

```ron
tabs: [
(
        name: "Queue",
        pane: Split(
            direction: Horizontal,
            panes: [
            (size: "40%", pane: Pane(Queue)),
            (
                    size: "60%",
                    pane: Pane(Property(
                        content: [
                        (kind: Property(Song(Title))),
                        (kind: Text(" Rating: ")),
                        (kind: Sticker("rating")),
                        ], align: Right,
                    )),
                ),
            ],
        ),
    ),
...
```

Go to said queue tab, press shift+R a few times on a few songs - note that the new property pane does not update the text.

Go to search tab and search for ratings in a way that makes them all show up, e.g. rating < 100

Only now are ratings up to date

mierak:
Ah got it, thank you very much. I definitely missed a spot (stickers appearing in property panes/header do not necessitate sticker fetching). I will try to come up with a more robust/generalized mechanism for this.

mierak:
Refreshing of the stickers should be hopefully solved once and for all.

Gremious:
Can confirm it works now, with the modal and the hotkey !

PR #685, opened 2025-10-18
feat: print config parse errors on commandline

### LINKED ISSUE #684: config errors hidden by MPD connection failure
## Bug Description

When rmpc fails to deserialize the config file due to syntax errors or invalid values, and MPD connection also fails, users only see the MPD connection error. The actual config deserialization error is hidden because the application exits before the UI initializes and displays the error modal.

This made it very difficult to debug config issues when first setting up rmpc, as the error message was completely misleading.

## Steps to Reproduce

1. Add an invalid action to your config (e.g., a non-existent keybind action like `VolumeSideways`)
2. Ensure you have a valid MPD socket configured in the config
3. Run `rmpc`

## Expected Behavior

User should see a clear error message indicating the config file has issues, with details about what went wrong and where.

## Actual Behavior

```
> rmpc
Error: Failed to connect to MPD

Caused by:
    GenericError: 'Connection refused (os error 111)'
```

The config deserialization error is completely hidden, and the error suggests an MPD problem even though the MPD address is configured correctly.

## Workaround

Running with an explicit address override allows the TUI to start and display the actual error:

```
> rmpc --address /home/simon/.mpd/socket
```

This shows a nice modal with the actual serialization error, making it possible to fix the config.

## Impact

This makes the initial configuration experience confusing and frustrating, as users are misled into thinking they have an MPD connection problem when the actual issue is in their config file.

404Simon:
## Description

Print config deserialization errors to stderr during startup to ensure users see them even when subsequent initialization steps (like MPD connection) fail.

Config errors were previously only displayed in a UI modal, which meant users never saw them if the application crashed during startup (e.g., MPD connection failure). This made debugging config issues difficult.

The fix adds `eprintln!("{err}")` to print config deserialization errors immediately to stderr, following the existing pattern used for other startup errors in the codebase. The error modal is still sent for display when the UI does initialize, preserving the hot-reload workflow.

### Related issues

Fixes #684 

### Checklist

- [x] All tests passed
- [x] Documentation has been updated (if needed) --> not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ remote commands inside scripts triggered by rmpc
 
 ### Fixed
 
+- Config deserialization errors are now printed to stderr during startup, making them visible even when subsequent initialization steps fail
 - Fixed `QueueTimeRemaining` not updating remaining time
 - `ModalClosed` event now correctly gets dispatched only after all modals were closed
 - Preview no longer disappears in search when returning to the search form while the results are


mierak:
Makes sense, thanks.

PR #689, opened 2025-10-19
update nord theme

v-Nyo:
## Description

Update if think it makes sense / like it?

Theming for new options and changed formatting.
Did not update the screenshot. Probably best that you test and add it if need be. I would wait to change it until rounded borders are added. 


mierak:
The theme works. Looks good, I like it more than before! I think its a good idea to merge this now, ratatui still has not released a new version.

The PR should get cleaned up first though, there is a bunch of commented out stuff that should be gone before it goes to the docs.

v-Nyo:
Thanks, Should I add image as well? (Concerned because git is no good with that) 
But if you want you could add your own or I could add this one?
<details>
rmpc
</details>

Are you going ahead with https://github.com/mierak/rmpc/pull/309?
If you do I would rewrite the theme before merging

mierak:
Sorry, I have been away from PC the last few days. The PR/theme looks good to me. Please do update the screenshot for now. I will look into a different solution for the images because I agree that the current state is less than ideal. After that we can merge this.

> Are you going ahead with https://github.com/mierak/rmpc/pull/309?
> If you do I would rewrite the theme before merging

Not for now, no. I would still like to get it in but am reconsidering if it is the right time for this change. I will probably update all the themes/docs myself if/when it gets in.

PR #691, opened 2025-10-23
Fix installation instructions for rmpc

sponkurtus2:
On the cargo install section, the command was wrong written

## Description
Word 'Install' was wrong

### Checklist

- [ Yes] All tests passed
- [ Yes] Code has been formatted with nightly
- [ Yes] Code has been checked with clippy (stable)
- [ Yes] Documentation has been updated (if needed)
- [ No] Changelog has been updated


mierak:
It is not wrong. [Cargo binstall](https://github.com/cargo-bins/cargo-binstall) is an existing tool which downloads precompiled binary from github releases. It is in the documentation in case someone does not want to wait for compilation and has cargo-binstall installed.

cargo install --locked and cargo binstall result in the same version being installed on your system.

PR #699, opened 2025-10-26
feat: add global crossfade control actions

404Simon:
## Description

- Adds global actions (`CrossfadeUp` and `CrossfadeDown` ) to control MPD crossfade ("xfade")
- Previously, crossfade could only be viewed via status properties - now it can be controlled interactively

### Related issues

None

### Checklist

- [x] All tests passed


mierak:
Thanks! Implementation in the MPD client is looking fine, but why are the controls in the outputs modal? Crossfade has nothing to do with outputs, this should be a global action similar to volume up/down.

404Simon:
> Thanks! Implementation in the MPD client is looking fine, but why are the controls in the outputs modal? Crossfade has nothing to do with outputs, this should be a global action similar to volume up/down.

You are right. I reverted the output modal changes and made it a global action.

mierak:
Needs docs update in the keybinds section https://mierak.github.io/rmpc/next/configuration/keybinds/#global. Just mentioning them is fine, don't set any default binds for them.

Other than that it looks good to me.

PR #735, opened 2025-11-19
feat: Hot-Reload Lyrics

MoreDelay:
## Description
I really like how lyrics are handled fully offline in rmpc, so I actually started to create my own synced lyrics. To do this, I basically edited the files manually and then tried to see the effect in rmpc. However, as the lyrics are only loaded at start-up, I had to restart the app a lot.

In this PR, I added a file watcher for the current lyrics in use. I basically copied the approach from the config hot-reload. Therefore I can image that you might want to move the `lyrics_watcher.rs` file somewhere else, or it might even be better to combine both file watchers to something more general. Tell me your thoughts. If you like these changes, I would also see to add the corresponding docs.

### Related issues
None

### Checklist

- [x] All tests passed

### DIFF
--- a/src/shared/lrc/index.rs
+++ b/src/shared/lrc/index.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::BTreeMap,
     io::BufReader,
     path::{Path, PathBuf},
     time::Duration,
@@ -24,7 +24,8 @@ use crate::{
 /// processing the entire content of each LRC file during startup.
 #[derive(Debug, Default, Serialize)]
 pub struct LrcIndex {
-    index: HashMap<PathBuf, LrcMetadata>,
+    // Using BTreeMap to have a well-defined iteration order.
+    index: BTreeMap<PathBuf, LrcMetadata>,
 }
 
 impl LrcIndex {
@@ -33,7 +34,7 @@ impl LrcIndex {
         let dir = WalkDir::new(lyrics_dir);
         log::info!(dir:?; "Starting lyrics index lyrics");
 
-        let mut index = HashMap::new();
+        let mut index = BTreeMap::new();
         for child in dir {
             let child = try_cont!(child, "skipping child");
             let child = child.path();
@@ -48,7 +49,7 @@ impl LrcIndex {
             log::trace!(child:?; "Successfully indexed lyrics file");
 
             {
-                use std::collections::hash_map::Entry;
+                use std::collections::btree_map::Entry;
                 match index.entry(child.to_path_buf()) {
                     Entry::Occupied(mut entry) => {
                         entry.insert(metadata);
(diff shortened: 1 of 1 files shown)


mierak:
Hi, hot reloading for lyrics is a fine feature imho, but I don't entirely agree with this approach.

We already have `rmpc remote indexlrc` on the CLI and can reuse its "machinery" to route the signal that the lyrics have changed from watcher to the pane with minimal effort by sending a `WorkRequest::IndexSingleLrc` message to the worker channel.

You are starting a recursive watcher on the whole parent directory of the lrc file anyway, so we might as well watch the whole `lyrics_dir` and react to any `*.lrc` changes and not have to deal with restarting the watcher all time.

All in all, just creating the lyrics watcher once, sending work request on every change (and deletion?) and doing some minor adjustments (mainly updating the LyricsIndexed event handler in the Lyrics pane and probably changing the index from Vec to a HashMap keyed by path so we do not get duplicate lrc entries) along with the config option you already have should be enough for this. I would also consider leaving the watcher off by default, not sure how many people need this, but that is not a big deal either way.

MoreDelay:
I reworked the feature based on your suggestions. Changing the `LrcIndex` to a `HashMap` resulted in a larger refactor than I originally anticipated. This also unveiled two bad tests. `shared::lrc::index::tests::song_has_no_duration` in particular is not a good test as there is no rule for what we expect to come out. I ignored it for now, but maybe we should delete it? The other test was `shared::lrc::index::tests::multiple_matches_no_lrc_without_len_no_length_match` where two entries had the same length difference to the target.

Furthermore, it appeared to me that `LrcIndexEntry` did not really fill a purpose as without the `path` field, it is just a subset of `LrcMetadata`, so I replaced it with that.

Feel free to leave further suggestions.

mierak:
> shared::lrc::index::tests::song_has_no_duration in particular is not a good test as there is no rule for what we expect to come out. I ignored it for now, but maybe we should delete it?

That test is/was fine imo, it specifically tests the unlikely(as I have mentioned in the comment) branch where we return the first result because there is really nothing better to do. This case not really being documented is ok, it shouldnt happen, but rmpc should behave consistently in the case it does.

> The other test was shared::lrc::index::tests::multiple_matches_no_lrc_without_len_no_length_match where two entries had the same length difference to the target.

This one was fine too I think, the case this tested was rmpc not crapping out when there are multiple lrc files with same length difference and that it returns a consistent result. I wouldn't touch this one.

These tests already did their job as well because they made us aware that the current implementation of these two cases is not stable (meaning two same lookups might not yield the same result due to hashmap not having a defined order). I think this should be remedied to still be stable so that users dont get inconsistent results if these cases ever happen.

As for the rest of the changes, I will take a closer look tomorrow since I have been quite busy the last few days but they look good from a cursory glance.

mierak:
looks good to me, thanks!

PR #754, opened 2025-11-30
feat: support abstract socket on Linux

szclsya:
## Description
This add support abstract socket (for Linux only). This is supported by MPD itself and ncmpcpp, and is pretty handy since user won't need to find a place to put the socket in. 

Linux itself identifies an abstract socket by a leading `\0`, but obviously this won't work for config files/env vars. This patch uses MPD's syntax (leading `@`). See https://mpd.readthedocs.io/en/latest/client.html#environment-variables.

### Related issues
N/A

### Checklist

- [x] All tests passed


mierak:
Thanks! macos build is failing, otherwise looks good to me just needs a [docs PR](https://github.com/rmpc-org/rmpc-org.github.io/blob/master/src/content/docs/next/configuration/index.mdx#address), should mention that the abstract socket is linux only too.

szclsya:
Done & done.

mierak:
Thanks!

Issue #723, opened 2025-11-11
Can't use arrows keys when editing new playlist name


Solarunit:
Testing new cool feature - new playlist name dialog in v0.10.0. 
You can't use arrows keys when editing new playlist name (only backspace is available) 

Also not possible activate Save/Cancel buttons with arrows.

mierak:
> You can't use arrows keys when editing new playlist name (only backspace is available)

All the input boxes in rmpc behave in this way, they are very simplistic and improving them is somewhere on my list.

> Also not possible activate Save/Cancel buttons with arrows.

Not exactly sure what you mean by this, they can be activated by the movement action keybinds, these keybinds are disabled while you are typing into the input.

Solarunit:
>Not exactly sure what you mean by this, they can be activated by the movement action keybinds, these keybinds are disabled while you are typing into the input.

Well, such dialog window style suggests that you could use Down key to select Save button and press Enter to confirm or select Cancel and press Enter to confirm. Sure you can ignore it and use Esc and Enter. Maybe it's okay how it is.

mierak:
Sorry, took my sweet time to respond. Anyway, these inputs are supposed to more or less mimic vim's modes. Once you focus the input so that you can type into it you enter an "insert" mode which you need to explicitly exit with escape action to be able to interact with other elements of the ui.

What I would agree with is that there should be some better indication of a user being in the "insert" mode instead of just the block symbol inside the input.

Solarunit:
Not sure about vim's mode, I don't use it (I prefer micro editor)

>What I would agree with is that there should be some better indication of a user being in the "insert" mode instead of just the block symbol inside the input.


Playlist name text or/and its frame could change color

PR #759, opened 2025-12-05
feat: bash/emacs style text inputs

mierak:
## Description

Consolidates text input handling throughout the application. Makes the input better by implementing similar set of actions to bash/emacs. The panes/modal can hook into this via context and implementing their respective trait functions.

* left/right/ctrl+b/ctr+f - move left/right
* alt+b/alt+f/ctrl+left/ctrl+right - move word left/right
* ctrl+a/ctrl+e - move to start/end
* ctrl+h/ctrl+d - delete left/right
* ctrl+u/ctrl+k - delete from cursor to start/end
* ctrl+w/alt+backspace/ctrl+d - delete word left/right

The implementation is unicode/graphemes aware too. It does not match the behavior exactly 1:1 but this can be changed later on if people insist on it. Word boundaries should be as per uax#29.

Still WIP.

### Related issues

https://github.com/mierak/rmpc/issues/723

### Checklist

- [ ] All tests passed

### DIFF
--- a/src/ui/modals/add_random_modal.rs
+++ b/src/ui/modals/add_random_modal.rs
@@ -311,6 +311,9 @@ impl Modal for AddRandomModal<'_> {
             MouseEventKind::LeftClick
                 if self.input_areas[InputAreas::Tag].contains(event.into()) =>
             {
+                if ctx.input.is_active(self.count_buffer_id) {
+                    ctx.input.normal_mode();
+                }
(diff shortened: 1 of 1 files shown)


mierak:
Ok, this was a doozy, way bigger than I expected.

Anyway, close to done if anybody feels like testing this, I probably broke a bunch of stuff that I still need to test/fix.

PR #763, opened 2025-12-09
Fix macOS build: Make Linux-only socket code conditional

hystericca:
Fixes build on macOS by making abstract socket code Linux-only

Abstract Unix sockets (SocketAddr::from_abstract_name) aren't available on macOS, causing compilation to fail. This PR adds proper platform conditionals so the code only compiles on Linux where it's supported.

Changes:
- Made SocketAddr import conditional with #[cfg(target_os = "linux")]
- Wrapped abstract socket logic in platform checks
- Prefixed unused addr parameters on non-Linux platforms

Tested on macOS with MPD + CoreAudio, everything works.

### DIFF
--- a/src/mpd/client.rs
+++ b/src/mpd/client.rs
@@ -1,14 +1,13 @@
 #[cfg(target_os = "linux")]
 use std::os::linux::net::SocketAddrExt;
+#[cfg(target_os = "linux")]
+use std::os::unix::net::SocketAddr;
 use std::{
     collections::HashSet,
     io::{BufRead, BufReader, Write},
     net::{Shutdown, TcpStream},
     os::unix::net::UnixStream,
 };
-
-#[cfg(target_os = "linux")]
-use std::os::unix::net::SocketAddr;
 
 use anyhow::Result;
 use log::debug;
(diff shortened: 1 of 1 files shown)


src/mpd/client.rs
```diff
@@ -212,10 +215,10 @@ impl<'name> Client<'name> {
         let mut stream = match &self.addr {
             MpdAddress::IpAndPort(addr) => TcpOrUnixStream::Tcp(TcpStream::connect(addr)?),
             MpdAddress::SocketPath(addr) => TcpOrUnixStream::Unix(UnixStream::connect(addr)?),
-            MpdAddress::AbstractSocket(addr) => {
+            MpdAddress::AbstractSocket(_addr) => {
```
mierak: same here

soifou:
> any idea why the build and tests for darwin ran fine in the PR that introduced this?

Maybe because the test only try with x86 arch and not arm?

FWIW, building with rust nightly on my MBP `aarch64-apple-darwin` is fine but it generates some warnings (this PR try to address):

```
warning: unused import: `SocketAddr`
 --> src/mpd/client.rs:7:21
  |
7 |     os::unix::net::{SocketAddr, UnixStream},
  |                     ^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default

warning: unused variable: `addr`
   --> src/mpd/client.rs:133:40
    |
133 |             MpdAddress::AbstractSocket(ref addr) => {
    |                                        ^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_addr`
    |
    = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default

warning: unused variable: `addr`
   --> src/mpd/client.rs:215:40
    |
215 |             MpdAddress::AbstractSocket(addr) => {
    |                                        ^^^^ help: if this is intentional, prefix it with an underscore: `_addr`

warning: `rmpc` (bin "rmpc") generated 3 warnings (run `cargo fix --bin "rmpc" -p rmpc` to apply 3 suggestions)
    Finished `release` profile [optimized] target(s) in 2m 46s
```

mierak:
> FWIW, building with rust nightly on my MBP aarch64-apple-darwin is fine but it generates some warnings (this PR try to address):

ah this makes sense, still I would prefer a different approach here (probably cfg away the whole match arm so its idiomatic for all target_os) though

> Maybe because the test only try with x86 arch and not arm?

I however doubt this though, I don't think target arch influences what is available in std and even the [`SocketAddr`](https://doc.rust-lang.org/stable/std/net/enum.SocketAddr.html) states that it is supposed to be a portable implementation.

`
SocketAddr is intended to be a portable representation of socket addresses and is likely not the same as the internal socket address type used by the target operating system’s API. Like all repr(Rust) structs, however, its exact layout remains undefined and should not be relied upon between builds.
`

mierak:
Anyway, merging this, I polished the implementation a bit. Still would like to find out what exact the build was failing for @gorewife though.

vkluna:
I think it may have been a false negative on my end? The first time I tried to run it I just got the warnings and wasn't able to get it running, so I made the changes and was able to run it without issues. However I just built the old version and it was able to run, just with the warnings. Sorry about that

PR #819, opened 2026-01-12
feat(theme): header column widths

JakeStanger:
## Description

This adds a new `header_column_widths` theme config option which allows controlling the width of the left, center and right columns in the header widget.

This is useful if for example you do not want to use the center column, or want to weight any of the columns differently (ie to be able to display longer song titles).

Example:

```ron
header_column_widths: [50, 0, 50],
```

This allows the full album title to be displayed:

image

Vs current build with default `[30, 40, 30]` split:

image

Let me know if you'd like me to update the changelog and docs too - I can address that in the next couple of days hopefully.

### Related issues

None

### Checklist

- [x] All tests passed

### DIFF
--- a/src/config/defaults.rs
+++ b/src/config/defaults.rs
@@ -30,6 +30,10 @@ pub fn default_column_widths() -> Vec<u16> {
     vec![20, 38, 42]
 }
 
+pub fn default_header_column_widths() -> Vec<u16> {
+    vec![30, 40, 30]
+}
+
 pub fn bool<const V: bool>() -> bool {
     V
 }
--- a/src/config/theme/mod.rs
+++ b/src/config/theme/mod.rs
@@ -69,6 +69,7 @@ pub struct UiConfig {
     pub scrollbar: Option<ScrollbarConfig>,
     pub song_table_format: Vec<SongTableColumn>,
     pub song_table_album_separator: AlbumSeparator,
+    pub header_column_widths: [u16; 3],
     pub header: HeaderConfig,
     #[debug("{}", default_album_art.len())]
     pub default_album_art: &'static [u8],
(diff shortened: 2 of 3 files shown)


JakeStanger:
As an aside, I think there's some fairly quick wins to reduce a lot of config code:

- Defaults can be derived by `#[serde(default)]` at the struct level instead of the field level, with a manual `Default` impl. Serde will automatically merge in missing fields from the default impl. This avoids the need to dupe everything with `#[default = "defaults::some_func]`.
- I reckon with some use of `#[serde(deserialize_with = "path")]`, the need for two separate `UiConfig` and `UiConfigFile` structs can be eliminated. Likely the same story for the main config.

If you'd like, I can have a look at those.

mierak:
Looks ok to me!

> Let me know if you'd like me to update the changelog and docs too - I can address that in the next couple of days hopefully.

No need actually, I have removed mentions of the Header pane from the docs as I would like to move away from this implementation in favor of custom layout as can be seen in the new [default theme](https://rmpc.mierak.dev/next/themes/) which I did yesterday.
The pane will stay in the code as its a fairly isolated piece of code and it does not make sense to me to break configs by removing it, so this change is fine if you use it.

> Defaults can be derived by #[serde(default)] at the struct level instead of the field level, with a manual Default impl. Serde will automatically merge in missing fields from the default impl. This avoids the need to dupe everything with #[default = "defaults::some_func].

Good point, I think I even used it somewhere.

> I reckon with some use of #[serde(deserialize_with = "path")], the need for two separate UiConfig and UiConfigFile structs can be eliminated. Likely the same story for the main config.

Probably, its even used in the cava theme. I am not completely sold on this one though, there are non trivial differences between the *File structs and their runtime counterparts. But it sure would improve some error messages.
https://github.com/mierak/rmpc/blob/019c768b53e84160951a6ba451f23025301aa01f/src/config/theme/cava.rs#L258-L287

Anyway, I am fine with ditching the whole defaults file and going for Default impls instead. For the deserialization, I *think* it would be better but would probably have to see the result.

PR #820, opened 2026-01-13
Fix duration format

### LINKED ISSUE #808: Duration in seconds in song info pane isn't ideal
### Description

Hello! 

In the info pane, is there a setting to tell rmpc to display the song duration in a more readable format than raw seconds? Screenshot of what I'm talking about (on the right next to `Duration`):

Image

I know the time is shown in the common h:mm:ss format in the top left corner for the currently playing song, but I'm wondering if I can change the time format shown for every song when browsing directories -- or if it's currently hard-coded as described above. To my untrained, not-a-Rustacean eye, it looks hard-coded: https://github.com/mierak/rmpc/blob/f3c6f11774c845e09c8815c4d9f3998ae23570e5/src/ui/panes/mod.rs#L395

### Proposed solution

I would propose to change that to the same format as the now playing song info in the top left corner. Or expose a user-configurable setting in case some folks like it the way it is now with the raw seconds value.

Gri-ffin:
## Description

I introduced a configurable duration formatting system and tried to preserve zero breaking changes for existing users. I centralized all duration-related logic into a single `DurationFormat` enum, which allowed me to replace the previously scattered DurationExt traits and ad-hoc match blocks across the UI panes.

To prevent any unexpected UI changes on update, I implemented a "smart fallback" mechanism: if a separator is detected in the user’s existing TUI configuration, the system automatically defaults to the Human format. This ensures that current setups stay exactly as they are.

I also added a translation layer as a custom format variant with its own placeholder syntax (such as %M and %S). This fully decouples the user configuration from the underlying duration library’s implementation details, meaning we can change our internal logic without breaking user configs (which can be changed easily without leaking implementation details to the user). users can define custom duration formats directly in their config using placeholders like %d or %D for days, %h or %H for hours, %m or %M for minutes, and %s or %S for seconds, where the uppercase variants provide zero-padding.

Currently I added this implementation to the `QueueTimeTotal` and `QueueTimeRemaining`, I run my own tests and added tests for the formatting, also the tests for `QueueTimeTotal` and `QueueTimeRemaining` passed which is a good sign, but might need more testing though before I update docs.

### Related issues

closes #808 

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ after starting a ytdlp download
 
 ### Changed
 
-- `duration_format` is now a string template and configurable. Use `"%m:%S"` for classic format, `"%M:%S"` for zero-padded minutes, or custom templates with tokens like `%d/%D` (days), `%h/%H` (hours), `%m/%M` (minutes), `%s/%S` (seconds), `%t` (total seconds)
+- `duration_format` is now a string template and configurable for song durations. Use `"%m:%S"` for classic format, `"%M:%S"` for zero-padded minutes, or custom templates with tokens like `%d/%D` (days), `%h/%H` (hours), `%m/%M` (minutes), `%s/%S` (seconds), `%t` (total seconds)
 - **Breaking** Keybinds now only override the defaults. Meaning your configured keybinds are combined
 - **Breaking** `show_song_table_header` has been removed. The queue header is now in a separate `QueueHeader`
 pane. You will need to update your config to include it in your tabs.


src/config/duration_format.rs
```diff
@@ -0,0 +1,113 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Clone, PartialEq, Eq, Hash, Serialize, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum DurationFormat {
```
mierak: Does this enum really make sense? Having sane defaults is nice, but this feel like this is overcomplicating things no Why not have the format as a simple String or some String newtype and default to the human/classic implementation?

Gri-ffin:
Hi,

> And lastly, we have a complete freedom with the format here and dont have to be inspired by the terse strftime syntax, we can even take look at [time-rs](https://time-rs.github.io/book/api/format-description.html) syntax for inspiration which looks a bit more pleasing the eye.

 I intentionally kept the formatting logic minimal since this is a terminal music player and the formatter is only meant to track and display durations. My goal was to cover the existing common cases (minutes/seconds, hours when needed, optional days) without introducing any unnecessary abstraction or some weird complex regex, since this is a music player and not a calendar app, we don't have the 'minutes vs. months' conflict. `time-rs` has a much more complex setting to work with, for rmpc, I don't think anyone would have queues with months of total time, so when a user sees `m`, he would directly assume its minutes even without looking at the docs, and in my opinion it's much more minimalist and shorter to type, but we can get @fabiendelpierre opinion since he opened the issue. 

>Having options for padded and non padded values is great. Does it make sense to be able to display the whole duration in seconds? Meaning 90 instead of 1:30?

great point, I agree that adding a way to display total seconds (e.g., 90s instead of 1:30) is a great utility for the player. I'll add a token for that. maybe a `t`, `T` for padded.

Also I don't like that `Human` field, I initially did it for testing and convention from **human readable**, but it's so vague and ambiguous, I thought why not name it `Default` or `Concise`, if you have any suggestion I am happy to listen.

I’ll hold off on the implementation until I hear your thoughts on these specifics, but I think this keeps the engine simple while giving the flexibility you asked for for future tokens.

mierak:
Apologies for leaving this hanging, I did not have access to my PC the past few days.

> I intentionally kept the formatting logic minimal since this is a terminal music player and the formatter is only meant to track and display durations.

Yeah I fully agree. I did not mean to implement some overengineered formatting logic for what this is. I am mostly ok with the provided implementation (only thing I am missing are the whole seconds, really). What I meant is that we do not have to keep this terse, strftime-like, syntax so that users dont need a lookup table when setting their custom format.

Imagine you are configuring rmpc and now you want your seconds to be padded, was it `%s` or `%S`?. Similar with users trying to display whole seconds, was it `%t`? `%f`? something else? Also if we went with `%T` for padded seconds only, what do we pad to exactly - 2 digits? 3? 4?

That is why I proposed taking inspiration from time-rs, its formatting syntax is pretty human readable in my opinion. For example, we could do: 
* `[second padding:zero]` (and `[second]` to default to padding with 0?) for `%S`
* `[second padding:none]` for `%s`
* `[seconds padding:zero digits:3]` instead of `%T` for example. It is verbose but more flexible and readable.

Where the padding can be zero/none/space and digits is a single whole number and possible set of values is [second]/[seconds]/[minute]/[hour]/[day] or something of this sort (basically the same what you have). And the `digits` is only relevant for the `%T` equivalent.

Nothing is set in stone, I am mostly thinking out loud, but this should not be that much more work that what you have already done, what do you think?

> Also I don't like that Human field, I initially did it for testing and convention from human readable, but it's so vague and ambiguous, I thought why not name it Default or Concise, if you have any suggestion I am happy to listen.

Honestly, I think getting rid of the enum completely might be the best option here. Falling back to the current behavior for QueueTimeTotal and Remaining is great, providing sane defaults for the rest can be done via the default config. The enum feels like it is overcomplicating this a bit.

Gri-ffin:
Hi,

>Honestly, I think getting rid of the enum completely might be the best option here. Falling back to the current behavior for QueueTimeTotal and Remaining is great, providing sane defaults for the rest can be done via the default config. The enum feels like it is overcomplicating this a bit.

sure, I'll go ahead and remove it.

but regarding the `[second padding:zero]` syntax, I'm still concerned it might be overkill for this use case. Implementing it properly would require a custom parser or regex, which adds (in my opinion) unnecessary complexity and potential for bugs and silent fails that would be nightmarish to debug for users, most CLI and terminal users are used to `strftime` syntax and its conventions, and I genuinely think simplicity is better here.

How about this: I'll use the verbose tokens (e.g., %minute, %Minute, %second, %Second) and add %total_sec.

While the capital letter doesn't explicitly say '2 digits,' it follows the common convention where the uppercase variant is the padded one. It's much easier to remember than %s vs %S. and the padding can be documented in the docs for reference.

If later users specifically need custom digit padding or space-padding, we can upgrade to a more complex parser then. But for now, this keeps the code extremely simple and ensures zero breakage for existing users.

mierak:
Lets go ahead with the strftime-like syntax then. Worst case it will have to sit alongside the more complex format in the future to not break existing configs.

> How about this: I'll use the verbose tokens (e.g., %minute, %Minute, %second, %Second) and add %total_sec.

That's not needed, at that point I would rather have the shorter syntax since as you mention it should feel familiar to users.
The %t and %T tokens for seconds only as you mentioned previously also seems fine.

Still, as I mentioned in the comments, I think this should be parsed once when reading the config into some intermediate representation and then simply stringified from there so we dont parse or repeatedly replace and allocate new strings on every render for every formatting token. This should be simple to write by hand or you can use the winnow which is in rmpc already if you want.

mierak:
Looks good, I realized/noticed one more thing just now though, sorry.

Consider the default config which uses `QueueTimeTotal()` meaning the separator is `None`. With a very long queue on current master it displays `5d 21:09:57`, with these changes it falls back to the `%m:%S` format so only `9:57` is displayed. Same goes for `QueueTimeRemaining`. This is not great.

Maybe `QueueTimeTotal` and `QueueTimeRemaining` should be special cased or not respect the new format at all and the old behavior kept?

Also #[allow(clippy:unwrap_used)] would be nice on the test module in duration_format.

Gri-ffin:
makes sense, I guess I can check for hours, days, minutes etc and manually build the string with either the custom separator or fall back to the old formatting if no separator specified for the Queues. Basically forrmatting would look like this:

- `1h 1m 1s` if no separator specified
- or `1h, 1m, 1s` if no separator specified

or

- `1h:1m:1s` if separator is `:`

it's best of both worlds?
I think it will also allow me to bring back the tests i removed for the old formatting:
```rust
#[test_case(Duration::from_secs(0), "0s")]
        #[test_case(Duration::from_secs(1), "1s")]
        #[test_case(Duration::from_secs(60), "1m")]
        #[test_case(Duration::from_secs(95), "1m, 35s")]
        #[test_case(Duration::from_secs(3600), "1h")]
        #[test_case(Duration::from_secs(3601), "1h, 1s")]
        #[test_case(Duration::from_secs(3661), "1h, 1m, 1s")]
        #[test_case(Duration::from_secs(3600 * 24), "1d")]
        #[test_case(Duration::from_secs(99999), "1d, 3h, 46m, 39s")]
        fn duration_format(input: Duration, expected: &str) {
            assert_eq!(input.format_to_duration(", "), expected);
        }
```

mierak:
Not sure I understand 100% correctly.

> * `1h 1m 1s` if no separator specified
> 
> * or `1h, 1m, 1s` if no separator specified

The second case is supposed to be if separator is `, ` right?

> or
> 
>     * `1h:1m:1s` if separator is `:`

That is the old behavior, same as the second case above with separator set to `:`, no?

I am starting to think that we shouldn't touch the current formatting in the QueueTime properties in this PR and keep the old behavior. Maybe in the future with changes to include some sort of trivial groupings in the new format along with deprecating the current separator field? Say `[%dd] %H:%M:%S` to keep the `5d 21:09:57` and to not display the days when they are 0.

Gri-ffin:
you're right, let's keep this simple and one at a time. I'll keep this pr only for the song duration.

mierak:
Thanks!

PR #821, opened 2026-01-13
refactor(config): use `#[serde(default)]` at struct level where possible

JakeStanger:
## Description

This is a first pass of the config code to remove a lot of the duplicated serde defaults code, deriving from the `Default` impl where possible (and providing one where sensible).

In some cases it didn't make sense as only a single field was being touched. Enum struct variants have also been left alone as this likely needs further refactoring (serde might let you derive it here, not sure...).

There are a few cases where this may change current functionality despite tests passing. I found several cases where the `#[serde(default)]` field attribute did not match the value provided in the `Default` implementation. Careful review of whether any of these have changed for the worse is needed, and I'll aim to do a review of this myself in the next few days.

Based on the above, I'll mark as draft for now.

I've left the separate `*File` structs for now, as I feel that's best tackled separately (and I'm yet to even check how feasible it is).

### Related issues

As discussed in #819 

### Checklist

- [x] All tests passed
- [-] [Documentation](https://github.com/rmpc-org/rmpc-org.github.io) has been updated (if needed)
- [-] Changelog has been updated

### DIFF
--- a/src/config/album_art.rs
+++ b/src/config/album_art.rs
@@ -4,20 +4,28 @@ use strum::Display;
 use super::Size;
 use crate::{mpd::mpd_client::AlbumArtOrder, shared::terminal::ImageBackend};
 
-#[derive(Debug, Default, Serialize, Deserialize, PartialEq, Eq, Clone)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+#[serde(default)]
 pub struct AlbumArtConfigFile {
-    #[serde(default)]
     pub method: ImageMethodFile,
-    #[serde(default)]
     pub order: AlbumArtOrderFile,
-    #[serde(default)]
     pub max_size_px: Size,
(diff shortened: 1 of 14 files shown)


JakeStanger:
> Do you know which these are?

There may be one or two I've missed, but I've identified the ones I found on review. Generally the discrepancies are between having no theme config vs having a minimal theme (ie removing most of the generated config), so I imagine most people have never spotted them. 

I think the scrollbar conflict resolved okay, but I'd just double check that to make sure I've not accidentally reverted them.

mierak:
I also found `browser_song_sort` which defaulted to `[Disc, Track, Artist, Title]` with no config and to empty array with config and when not specified, that was a bug which this fixes.

The other properties not having to be specified anymore and falling back to default is good too. They were mostly leftovers from the early mostly which I never bothered to change. The progress bar looks good too.

Great job, thank you!

PR #841, opened 2026-01-28
Fix changelog

tremby:
Looks like an entry was partially moved.


tremby:
Looking again, it seems there are more issues.

Example: the line `with the default ones. Set ``clear`` to true to keep the old behavior.` -- I'm not sure what that's a continuation of.

tremby:
Oh, it looks like it goes after the breaking change re keybinds. Fixing and updating commit.

mierak:
How did I even manage that, thanks!

PR #858, opened 2026-02-04
Add nix flake

Kladki:
## Description

This flake mainly serves the purpose of allowing people to install the latest git version of rmpc
easily, with having things like shell completions installed by default, as well as providing
another method to try it out without installing.

In addition, it also provides a development shell, so that the rust toolchain is available
immediately on nix-enabled systems. If rmpc ever adds any system dependencies in the future, these
can be added too.

As I understand how intimidating nix can be at first, I made sure to add lots of comments to explain
what is going on, as well as keeping the flake relatively simple, not doing anything complicated like
cross-compilation or anything that will need active maintenance. The only thing is updating the rust
toolchain to the latest version if the MSRV goes above what nixpkgs had at this time (1.92), which
can be done easily with `nix flake update`.

On top of that, you can always mention me if any issues ever arise.

### Checklist

- [x] All tests passed

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 /target
 **.ron
 !assets/**.ron
+# Nix
+/.direnv
+result
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,122 @@
+{
+  "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "naersk",
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1752475459,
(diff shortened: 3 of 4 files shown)


mierak:
Right, so if something changes about completions etc it will need to be updated, I will try to keep it in mind.

Also, regarding the rust version, I already wait a while if I want to update MSRV in rmpc because users then report that they are not able to compile rmpc and are not realizing that they are on ae older toolchain version. I assume a good practice from now on would be to check nixpgks unstable if that rust version is available in there?

Kladki:
> Also, regarding the rust version, I already wait a while if I want to update MSRV in rmpc because users then report that they are not able to compile rmpc and are not realizing that they are on ae older toolchain version.

Ah alright, that's good. In light of that, this:

> I assume a good practice from now on would be to check nixpgks unstable if that rust version is available in there?

Shouldn't really a problem, since I don't think nixpkgs would lag more than 1 version behind the latest release. I only think that 1.93.0 is not available yet as it seems all rust packages must be rebuilt before the new version is available. If that ever becomes a problem, we can just get the rust toolchain from a rust-specific source, like [fenix](https://github.com/nix-community/fenix), which is updated daily for stable and nightly releases.

PR #869, opened 2026-02-13
Resolve environment variables when parsing paths from the config file

UnnaturalTwilight:
## Description
Add vary simple environment variable substitution, similar to the tilde expansion, when parsing paths in the config file.
This needs more testing, especially with regards to external commands.

The environment variable substitution isn't run or options from the cli or `MPD_HOST`/`MPD_PORT` variables as the shell will have already done so.

This also adds a check that the cache dir is an absolute path as I found while testing that the yt-dlp integration breaks if it isn't.

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ keybinds if they contained either `{` or `}`. You will now need to escape these 
 ### Changed
 
 - Improved error message when invalid `default_album_art_path` path is provided
-- Environment variables are not resolved when parsing paths in the config file or theme
+- Environment variables are now resolved when parsing paths in the config file or theme
 
 ### Fixed
 
--- a/src/config/album_art.rs
+++ b/src/config/album_art.rs
@@ -5,7 +5,7 @@ use strum::Display;
 
 use super::Size;
 use crate::{
-    config::utils::{env_var_expand, tilde_expand},
+    config::utils::tilde_expand,
     mpd::mpd_client::AlbumArtOrder,
     shared::terminal::ImageBackend,
 };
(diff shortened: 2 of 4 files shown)


UnnaturalTwilight:
I also found while testing this that the `default_album_art_path` option only works when specified in the theme, despite the docs saying it can be set in ether the config or theme.

mierak (commented):
Thanks for PR!

* The TODOs should probably get cleaned up, but I suppose thats TBD still since you marked it as a draft.
* The expansion is still missing in cli_config.rs, its a slimmed down version of the config file used only for the CLI commands.
* I am not completely sure on the expansion in commands, maybe only the first element in the array (the program) should get expanded? Users can already opt in to expansion by executing stuff through the shell: `["sh", "-c", "<command>"]`

> I also found while testing this that the default_album_art_path option only works when specified in the theme, despite the docs saying it can be set in ether the config or theme.

No idea what that is doing there, huh, it can be deleted from the config section.

UnnaturalTwilight:
I made the config error when cache_dir is invalid to the main config, however I only made it a warning in the cli config seeing as only the yt-dlp options require it. If you think it would be better to fully throw an error there it is an easy change.

For the various external command options I'm thinking it will probably be best to not expand env vars as any invoked shell will do a better job at parsing them correctly than what I implemented.

mierak:
> I made the config error when cache_dir is invalid to the main config, however I only made it a warning in the cli config seeing as only the yt-dlp options require it. If you think it would be better to fully throw an error there it is an easy change.

That sounds reasonable.

> For the various external command options I'm thinking it will probably be best to not expand env vars as any invoked shell will do a better job at parsing them correctly than what I implemented.

Leaving the commands in the same state as they were now is also fine by me.

mierak:
thanks!

PR #880, opened 2026-02-18
feat: add stop_on_exit config option

higorprado:
## Summary

- Adds a new `stop_on_exit` config option (default: `false`)
- When enabled, rmpc sends a stop command to MPD before terminating
- Allows users who prefer music to stop when closing rmpc to configure this behavior

## Usage

Add to your `config.ron`:
```ron
stop_on_exit: true,
```

## Test plan

- [x] Build and run with default config (`stop_on_exit: false`) - music continues playing after quit
- [x] Build and run with `stop_on_exit: true` - music stops when rmpc exits
- [x] Test quitting when music is paused - stops playback correctly
- [x] Test quitting when music is already stopped - no error
- [x] All 952 tests pass
- [x] Clippy passes with no warnings
- [x] Code formatted with `cargo +nightly fmt`
- [x] Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ keybinds if they contained either `{` or `}`. You will now need to escape these 
 - `scroll_speed` to `song_table_format`
 - `album_art.custom_loader` to allow for more flexibility when choosing the album art image
 - added `CopyToClipboard()` action
-- `stop_on_exit` config option to automatically stop playback when exiting rmpc
+- `on_exit` config option to execute a command when exiting rmpc, similar to `on_song_change`
 
 ### Changed
 
--- a/assets/example_config.ron
+++ b/assets/example_config.ron
@@ -20,7 +20,7 @@
     keep_state_on_song_change: true,
     reflect_changes_to_playlist: false,
     select_current_song_on_change: false,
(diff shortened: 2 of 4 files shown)


mierak:
Hi, I agree that this is useful, however I do not agree with the way its handled.

I think a much better and more flexible solution is to provide another hook allowing users to execute an arbitrary script on exit, similar to on song change. This would enable users to have any kind of behavior on exit that they might desire, be it a simple pause instead of stop, clearing the queue or what have you.

higorprado:
Thank you for the feedback! You're absolutely right - a hook is much more flexible.

I've refactored the implementation to use an on_exit hook instead of the boolean stop_on_exit option. This follows the same pattern as on_song_change and allows users to execute any command when exiting rmpc.

The hook receives the same environment variables as on_song_change (CURRENT_SONG, STATE, PID, etc.).

mierak:
Docs PR is needed as well

higorprado:
Thanks for the review! I've addressed both points:

1. Tilde expansion - Fixed in src/config/mod.rs:382, now follows the same pattern as on_song_change and on_resize.
2. Docs PR - Created at https://github.com/rmpc-org/rmpc-org.github.io/pull/39

PR #893, opened 2026-02-25
feat: support adding/removing songs to/from multiple playlists at once

techmoocher:
## Summary

Adds multi-select support to the save-to-playlist and delete-from-playlist modals, allowing users to add or remove songs to/from multiple playlists in a single operation.

## Changes

- **New `MultiSelectSection` widget** (`multi_select_section.rs`): A new modal section type that renders playlist items with `[x]`/`[ ]` checkboxes. Users toggle items using the `Select` keybind (Space by default), then press Confirm (Enter) to apply the operation to all checked playlists.
- **`Section` trait**: Added a `toggle()` method (default no-op) to support checkbox toggling.
- **`MenuModal`**: Wired `CommonAction::Select` to call `toggle()` on the active section. Added `multi_select_section()` builder method.
- **Save modal**: Now uses `MultiSelectSection` instead of `SelectSection`, enabling multi-playlist song addition with per-playlist duplicate detection.
- **Delete modal**: Also uses `MultiSelectSection`, enabling multi-playlist song removal.
- **Backward compatible**: If no items are explicitly checked, the modal falls back to single-select behavior (the highlighted item is used), preserving the existing UX for users who don't need multi-select.

## Motivation

Addresses the feature request: *"Would be nice if you could add and remove songs from multiple playlists at the same time."*

## Testing

- `cargo clippy` passes (stable)
- `cargo test` passes (all 952 tests)
- Formatted with `cargo +nightly fmt --all`

### DIFF
--- a/src/ui/modals/menu/mod.rs
+++ b/src/ui/modals/menu/mod.rs
@@ -42,6 +42,7 @@ mod input_section;
 mod list_section;
 pub mod modal;
 mod multi_action_section;
+mod multi_select_section;
 mod select_section;
 
 trait Section {
@@ -57,6 +58,9 @@ trait Section {
     fn select(&mut self, idx: usize);
     fn unselect(&mut self, ctx: &Ctx);
     fn unfocus(&mut self, _ctx: &Ctx) {}
+
+    /// Toggles the current item in multi-select sections. No-op by default.
+    fn toggle(&mut self) {}
 
     /// Returns true if the modal should be destroyed upon executing
     fn confirm(&mut self, ctx: &Ctx) -> Result<bool>;
@@ -80,6 +84,7 @@ enum SectionType<'a> {
     Select(SelectSection),
     Multi(MultiActionSection<'a>),
     Input(InputSection<'a>),
+    MultiSelect(multi_select_section::MultiSelectSection),
 }
 
 impl Section for SectionType<'_> {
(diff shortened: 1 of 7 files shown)


src/ui/modals/menu/mod.rs
```diff
@@ -426,55 +462,64 @@ pub fn create_save_modal<'a>(
...
-        all_songs.iter().cloned().partition(|s| songs_in_playlist.contains(s));
+pub fn add_to_playlists_or_show_modal(
+    playlist_names: Vec<String>,
+    all_songs: &[String],
+    duplicate_strategy: DuplicateStrategy,
+    ctx: &Ctx,
+) {
+    for playlist_name in playlist_names {
+        let pl_name = playlist_name.clone();
+        let all_songs = all_songs.to_owned();
```
mierak: This clones all songs for every single modal in question, would be nice to not have to do that.

mierak:
> New MultiSelectSection widget (multi_select_section.rs)

> Backward compatible: If no items are explicitly checked, the modal falls back to single-select behavior

Why is a new section type necessary? Why not edit the current one if the new one is backwards compatible with the existing behavior anyway?

> that renders playlist items with [x]/[ ] checkboxes.

This is not consistent with the rest of rmpc. There are symbols in config.

> cargo clippy passes (stable)

It does not. The CI is failing.

Additionally:

The UX side of things is not great. This displays up to N modals right after each other when the duplicate strategy is set to `Ask` with N being the number of selected playlists.
Selecting an item does not move to the next one like it does everywhere else in rmpc.
The notifications after saving/deleting are not great, it will show all status notifications in quick sequence so users will only ever see the last one.

I understand that this is most likely AI generated code, you should really scrutinize more thoroughly what it decides to generate.

PR #933, opened 2026-03-16
chore(flake): refactor, add rmpcd package

Kladki:
## Description

Switched from using naersk to crane, as it has better support for workspaces.
It also doesn't have that bug causing the `unused_crate_dependencies` to fail, which is a nice plus.

I also added the `rmpcd` binary to the flake, so that can be used as well.

Since `rmpcd` doesn't have any documentation on the website from what I've seen, I haven't added any for this PR either.

### Checklist

- [x] All tests passed

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,24 +1,17 @@
 {
   "nodes": {
-    "fenix": {
-      "inputs": {
-        "nixpkgs": [
-          "naersk",
-          "nixpkgs"
-        ],
-        "rust-analyzer-src": "rust-analyzer-src"
-      },
+    "crane": {
       "locked": {
-        "lastModified": 1752475459,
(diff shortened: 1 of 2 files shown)


mierak:
Thanks!

> Since rmpcd doesn't have any documentation on the website from what I've seen, I haven't added any for this PR either.

Yup, its still pretty early for rmpcd, its ~2 weeks old and still changing. But it is at a point where documentation is on the table, I will get to it sooner rather than later.

mierak:
Also on that note, if you have a use for it and find any issues(be it with rmpcd itself or with the lua API), dont hesitate to throw them at my face.

Kladki:
I probably won't find any issues haha, I'm just using it for the lyrics script to replace the bash script (and the play counter, since it's available why not?). As for the other features, I already use mpdris2-rs for mpris and listenbrainz-mpd for my scrobbling, while i don't feel the need to have notifications for song changes.

PR #934, opened 2026-03-16
fix(flake): re-add rust-analyzer

Kladki:
## Description

Just found out that it isn't included in the default shell, probably because they assume it'd already be installed system-wide, and it's version doesn't need to be in sync with the project.

Unfortunately I often only notice stuff like this a minute after changes have been merged. 😅

### Checklist

- [ ] All tests passed


mierak:
> Unfortunately I often only notice stuff like this a minute after changes have been merged. 😅

I do the same :smile: most of my pull requests have amended commits...

PR #935, opened 2026-03-16
feat: add inverted_bar_as_mask config option

Everesh:
## Description
- Adds a new `inverted_bar_as_mask` config option for cava, defaulting to false

Mostly intended to circumvent the legacy top aligned block chars, since they can be messy on some fonts. In my case `DejaVuSansM Nerd Mono` didn't play well with it.

|legacy top aligned block chars| masked out bottom aligned block chars |
|-|-|
| image | image |

This implementation should be fine even if there is no explicit bg set, since `.reverse()` is just a `\x1b[7m` prefix, but it will clash with the opacity of terminal emulators.

| terminal emulator opacity | WM opacity |
|-|-|
| image | image |

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ keybinds if they contained either `{` or `}`. You will now need to escape these 
 - Added `quit_closes_modal`
 - `remove` to the CLI allowing you to remove songs from the queue
 - `Pause` and `Unpause` actions
+- `inverted_bar_as_mask` option to cava
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


mierak:
These font issues are really a pain, I am fine with this option since its off by default.

Peknej theme, na prvni pohled jsem myslel ze to je ncmpcpp kvuli barvam na vizualizeru.

Everesh:
Díky, nebudu lhát, byla to pro mě nemalá inspirace.

PR #953, opened 2026-03-29
fix: recognize 'all' in allow-passthrough and detect iTerm.app

chadchae:
## Summary
- `is_passthrough_enabled()` only checked `.ends_with("on")`, missing tmux's `allow-passthrough all` setting. Added `"all"` as a valid value.
- `Emulator::detect()` TERM_PROGRAM fallback did not include `"iTerm.app"` or `"iTerm2"`, causing iTerm2 to go undetected when DA1 query fails (e.g., inside tmux with passthrough issues).

## Context
When running rmpc inside tmux on iTerm2:
1. `tmux show -Ap allow-passthrough` returns `"all"` by default, but `is_passthrough_enabled()` only matches `"on"` — users must explicitly set `on` instead of `all` as a workaround.
2. Inside tmux, the DA1 query (`\x1b[>q`) often fails to reach the outer terminal. The TERM_PROGRAM fallback then checks env vars, but `"iTerm.app"` (the value iTerm2 sets) was not in the match list, so iTerm2 goes undetected.

## Test plan
- [ ] Verify `is_passthrough_enabled()` returns `true` when `allow-passthrough` is set to `all`
- [ ] Verify `Emulator::detect()` correctly identifies iTerm2 via `TERM_PROGRAM=iTerm.app` fallback
- [ ] Verify album art displays correctly in tmux on iTerm2

🗿 MoAI <email@mo.ai.kr>

### DIFF
--- a/rmpc/src/shared/terminal/emulator.rs
+++ b/rmpc/src/shared/terminal/emulator.rs
@@ -59,6 +59,7 @@ impl Emulator {
             "WezTerm" => Some(Emulator::WezTerm),
             "vscode" => Some(Emulator::VSCode),
             "Tabby" => Some(Emulator::Tabby),
+            "iTerm.app" | "iTerm2" => Some(Emulator::Iterm2),
             _ => None,
         };
 
--- a/rmpc/src/shared/tmux.rs
+++ b/rmpc/src/shared/tmux.rs
@@ -154,7 +154,8 @@ pub fn is_passthrough_enabled() -> anyhow::Result<bool> {
     let cmd = cmd.args(["show", "-Ap", "allow-passthrough"]);
     let stdout = cmd.output()?.stdout;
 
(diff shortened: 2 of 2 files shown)


mierak:
Hi, thanks for the contribution. Please use the pull request template as it mentions important steps for pull requests.

The change to include "all" in allow-passthrough looks ok.

Regarding the TERM_PROGRAM, looking at iterm2 source code I see no mention of the "iTerm2" value, where is this coming from? https://github.com/search?q=repo%3Agnachman%2FiTerm2%20TERM_PROGRAM&type=code

Changelog is missing.

mierak:
Hi, do you still want to follow up on the points I raised?

mierak:
finished in https://github.com/mierak/rmpc/pull/973

Issue #920, opened 2026-03-12
Better `text_color`


VigneshR387:
### Description

The current `text_color` have little to no visual impact on the theme. This is happens because because each `pane(Property())` & `Song_table_format` follows their own custom `style(fg:" ")`. If not defined, they default to white as their foreground  for text (I might be wrong here). This makes it a hassle to set to a uniform text color that spans across the entire theme.

**Example:**
```
    background_color: "white",
    text_color:"black" ,
```

<img width="1862" height="975" alt="Image" src="https://github.com/user-attachments/assets/2e344ef5-58d1-4916-ab7c-df591daf4fd9" />

### Proposed solution

Make `text_color` override the foreground color  for `Pane(Property())`  and `Song_table_format` unless each pane explicitly set the `style(fg: "")` . This way if  the user wants to set a custom foreground color for a specific pane, they can do so.

mierak:
mostly fixed in https://github.com/mierak/rmpc/pull/981

PR #979, opened 2026-04-25
apply `text_color` to more elements

Undeadamien:
## Description
This PR applies `text_color` to more elements.
I also added a test theme with relevant `fg:` values removed to make verification easier. It should be removed before merging.

### Related issues
https://github.com/mierak/rmpc/issues/920

### Checklist

- [X] All tests passed
- [X] Code has been formatted with rustfmt (nightly)
- [X] Code has been checked with clippy (stable)
- [X] Changelog has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to this project will be documented in this file.
 
 - Improved error message when invalid `default_album_art_path` path is provided
 - Environment variables are now resolved when parsing paths in the config file or theme
+- Apply `text_color` to more elements
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


mierak:
Hi, thank you for your contribution and I am extremely sorry for the very late reply. I have not had time to properly look at this and was doing just basic maintenance/small changes, especially since https://github.com/mierak/rmpc/pull/981 appeared at basically the same time solving the same problem.

Looking at both of these, I have to decide for one or the other so I will close this in favor of #981 as it seems a bit more complete.
Again, I am extremely sorry for this, thank you!

PR #981, opened 2026-04-25
Improve styling 

UnnaturalTwilight:
## Description

This makes more styles stack on others to make theming more consistent.
- 'current' and 'highlight' versions of symbols styles stack on the base style for that symbol
- Models now use `text_color`
- `highlight_border_style` defaults to `borders_style`
- tab bar defaults to using `current_item_style` for active tab and `text_color` for inactive tab

Only the model change should affect fully defined themes but the rest of the changes make theming more forgiving to incomplete themes and reduce the effects of adding new theme options as they inherit from the existing options.

### Related issues

This is progress on #920 but not a full implementation of text color

### Checklist

- [x] All tests passed

If these changes are ok I will work on a pr for the docs.

### DIFF
--- a/rmpc/src/config/theme/mod.rs
+++ b/rmpc/src/config/theme/mod.rs
@@ -437,9 +437,11 @@ impl TryFrom<UiConfigFile> for UiConfig {
         let fg_color = StringColor(value.text_color.clone()).to_color()?;
         let bg_color = StringColor(value.background_color).to_color()?;
         let header_bg_color = StringColor(value.header_background_color).to_color()?.or(bg_color);
-        let fallback_border_fg = Color::White;
+        let border = value.borders_style.to_config_or(Some(Color::White), None)?;
         let border_set_lib: BorderSetLib = value.border_symbol_sets.try_into()?;
         let components = convert_components(value.components, &border_set_lib)?;
+        let current_style = value.current_item_style.to_config_or(None, None)?;
+        let highlighted_style = value.highlighted_item_style.to_config_or(None, None)?;
 
         Ok(Self {
             layout: value.layout.convert(&components, &border_set_lib)?,
@@ -455,12 +457,15 @@ impl TryFrom<UiConfigFile> for UiConfig {
             modal_backdrop: value.modal_backdrop,
             text_color: fg_color,
             header_background_color: header_bg_color,
-            borders_style: value.borders_style.to_config_or(Some(fallback_border_fg), None)?,
+            borders_style: border,
             highlight_border_style: value
                 .highlight_border_style
-                .to_config_or(Some(Color::Blue), None)?,
+                .to_config_or(border.fg, border.bg)?,
             symbols: value.symbols.into(),
-            scrollbar: value.scrollbar.map(|sc| sc.into_config(fallback_border_fg)).transpose()?,
+            scrollbar: value
+                .scrollbar
+                .map(|sc| sc.into_config(border.fg.unwrap_or(Color::White)))
+                .transpose()?,
             progress_bar: value.progress_bar.into_config()?,
             song_table_format: TryInto::<QueueTableColumns>::try_into(value.song_table_format)?.0,
             song_table_album_separator: value.song_table_album_separator,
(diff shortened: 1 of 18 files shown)


rmpc/src/config/theme/mod.rs
```diff
@@ -434,11 +434,14 @@ impl TryFrom<UiConfigFile> for UiConfig {
 
     #[allow(clippy::similar_names)]
     fn try_from(value: UiConfigFile) -> Result<Self, Self::Error> {
+        let fg_color = StringColor(value.text_color.clone()).to_color()?;
         let bg_color = StringColor(value.background_color).to_color()?;
         let header_bg_color = StringColor(value.header_background_color).to_color()?.or(bg_color);
-        let fallback_border_fg = Color::White;
+        let border = value.borders_style.to_config_or(Some(Color::White), None)?;
```
mierak: Mostly just thinking, maybe its time to get rid of the White color? I think these defaults were a mistake on my part back then and these defaults should come from the ConfigFile::default() if anything instead. This default essentially forbids users from unsetting the border style completely. Its used in many places so a pass on this could be done in the future/another PR.

rmpc/src/ui/modals/input_modal.rs
```diff
@@ -146,9 +146,10 @@ impl<'a, C: FnOnce(&Ctx, &str) -> Result<()> + 'a> Modal for InputModal<'a, C> {
             .build();
 
         self.button_group.set_active_style(if ctx.input.is_active(self.input_buffer_id) {
+            // !!TODO - Should this have config??
```
mierak: Isn't there some sane combination that we can come up with? What about
```rust
Style::default()
    .fg(ctx.config.theme.background_color.unwrap_or_default())
    .bg(ctx.config.theme.text_color.unwrap_or_default())
```
This should ensure that there is always contrast between the colors when an input is active and it should also be distinct from the current item style unless the user sets it otherwise, what do you think?

mierak (commented):
Sorry I took my sweet time, dont really have that much time lately.

I always have hard time wrapping my head around these styling changes for some reason but unless I am missing something, this seems mostly OK to me.

> 'current' and 'highlight' versions of symbols styles stack on the base style for that symbol

Is this really wanted though? Doesn't this just limit user's choices? If they want underline for the normal marker style for whatever reason but do not want it for the highlight style this will essentially disallow that no? Then again it might be more consistent with the other styles...

Probably a few forgotten things:
- ~~Tabs pane should probably merge on top of text style too to be consistent right?~~
- `as_line()` is done for the queue table etc but `as_span` is not so properties wont inherit the text style still

UnnaturalTwilight:
> Is this really wanted though? Doesn't this just limit user's choices? If they want underline for the normal marker style for whatever reason but do not want it for the highlight style this will essentially disallow that no? Then again it might be more consistent with the other styles...

I agree that it is not a great limitation, but it only applies to modifiers given ratatui doesn't have a explicit value representing no modifiers that can be set in the config where as the colours can be set to `default`. I think it is an acceptable trade off for now for consistency and backwards compatibility but understand if you would rather hold off on it.

mierak:
Lets keep it I suppose, the colors are always overrideable as you said yourself so its just the modifiers and these cases where it would affect somebody are probably pretty rare.

If the need comes up, we might be able to expose ratatui's [sub_modifier](https://docs.rs/ratatui/latest/ratatui/style/struct.Style.html#structfield.sub_modifier) which should make removing modifiers possible I think.

rmpc/src/ui/modals/input_modal.rs
UnnaturalTwilight: Using `highlighted_item_style` would make more sense to me but im unsure if that would be distinct enough from current item on all configs

rmpc/src/config/theme/mod.rs
UnnaturalTwilight: I got rid of the ones I could find with a quick search

rmpc/src/ui/modals/input_modal.rs
mierak: It should be, otherwise users will not see the current song in queue or their filtered items etc.

mierak:
Seems ok, thanks!

PR #1006, opened 2026-05-08
Fix fg not applying in `current_item_style`and `highlighted_item_style` in `Reversed` mode

### LINKED ISSUE #1003: fg does not apply in current_item_style
### Description

Image

### Steps to reproduce

`current_item_style: (fg: "red", bg: "#81a1c1", modifiers: "Bold | Reversed"),`

Gri-ffin:
## Description

fixes fg not applying in current_item_style

### Related issues

closes #1003

### Checklist

- [x] All tests passed


mierak:
Thanks!
While testing this I also found that the same bug exists for `highlighted_item_style`. I think that applying the same fix should be applied to both the styles in QueueRow.

Gri-ffin:
now both should work.

mierak:
thanks!

PR #1016, opened 2026-05-23
add a vertical slider for volume

### LINKED ISSUE #1008: Vertical volume slider
### Description

I have a mostly vertical arrangement for my RMPC theme, and I'd like to be able to use a vertical volume slider to go along with it. That option doesn't exist right now, and the existing volume slider also doesn't wrap to the next available line, so there's no way to achieve it with theme settings. I'd appreciate if a vertical volume panel was implemented either as an alternative or as a configuration option for the existing volume slider. Thanks!

### Proposed solution

_No response_

Gri-ffin:
## Description

adds a new vertical slider pane to control the volume, [link](https://github.com/rmpc-org/rmpc-org.github.io/pull/65) for doc pr.

### Related issues

closes #1008 

### Checklist

- [x] All tests passed

### DIFF
--- a/rmpc/src/ui/panes/volume.rs
+++ b/rmpc/src/ui/panes/volume.rs
@@ -67,6 +67,44 @@ impl Pane for VolumePane {
                     frame.buffer_mut().set_string(area.x + i, area.y, c, style);
                 }
             }
+            VolumeType::VerticalSlider(config) => {
+                if area.height < 1 || area.width < 1 {
+                    return Ok(());
+                }
+
+                let symbols = &config.symbols;
+                let filled_len = (f64::from(area.height - 1)
+                    * f64::from(*ctx.status.volume.value())
+                    / 100.0) as u16;
+
+                for i in 0..area.height {
+                    let inv_i = area.height - 1 - i;
+
+                    let style = if inv_i <= filled_len && filled_len > 0 {
+                        config.filled_style
+                    } else {
+                        config.track_style
+                    };
+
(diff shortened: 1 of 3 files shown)


rmpc/src/ui/panes/volume.rs
```diff
@@ -87,9 +125,22 @@ impl Pane for VolumePane {
...
                 }
 
-                let volume_ratio =
-                    f32::from(event.x.saturating_sub(self.area.x)) / f32::from(self.area.width - 1);
-
+                let volume_ratio = match &self.config {
+                    VolumeType::Slider(_) => {
+                        if self.area.width == 0 {
+                            return Ok(());
+                        }
```
mierak: This check is unnecessary no? It happens just a few lines above.

rmpc/src/ui/panes/volume.rs
```diff
@@ -87,9 +125,22 @@ impl Pane for VolumePane {
...
+                        f32::from(event.x.saturating_sub(self.area.x))
+                            / f32::from(self.area.width - 1)
+                    }
+                    VolumeType::VerticalSlider(_) => {
+                        if self.area.height == 0 {
```
mierak: same here

mierak (commented):
Thanks! Save for the small duplicated check this looks good to me.

rmpc/src/ui/panes/volume.rs
Gri-ffin: good catch completely missed that

mierak:
thanks!

PR #1021, opened 2026-05-28
fix(rmpcd): emit PropertiesChanged for Can* properties

cwelsys:
## Description
notify.rs doesn't emit PropertiesChanged for CanPlay/CanGoNext/CanGoPrevious, so clients that cache them never see them update. Emit them alongside the playback-state, metadata and queue changes they derive from.

### Checklist

- [x] All tests passed


mierak:
looks ok to me, thanks!

PR #1022, opened 2026-05-29
feat: rmpcd - implement mpris Seek, SetPosition and Seeked

cwelsys:
## Description
- `Seek` was treating its offset as an absolute position; it now seeks relative to the current position.
- `SetPosition` does an absolute seek, ignoring out-of-range positions or a track ID that isn't the current song.
- `Seeked` fires after a successful seek.

### Checklist

- [x] All tests passed


mierak:
Sorry for the hold up, seems to be up to spec so looks good to me, thanks!

Note: the CI failures are not this PR's fault

PR #1030, opened 2026-06-16
feat(queue): add Vim-style focus entry positioning keybinds

trumpeter122:
## Description

Added three Vim motion keybinds, `zt`, `zz`, and `zb`, to scroll focused entry in a queue to the top, middle, or the bottom of the viewport.

### Related issues

None.

### Checklist

- [x] All tests passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3346,6 +3346,7 @@ dependencies = [
  "clap",
  "ctrlc",
  "derive_more",
+ "log",
  "md5",
  "mlua",
  "reqwest",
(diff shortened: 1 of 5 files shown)


mierak:
Thanks! Looks mostly alright, but:

* There is no reason for this to be under queue actions, it is already implemented on the dir, moving this to navigation keybinds should be simple and will allow this to work in more places like the browser panes
* This should respect [scrolloff](https://rmpc.mierak.dev/configuration/#scrolloff) for consistency. The cursor jumps with a non zero scrolloff when moving after using `zt`/`zb`
* The documentation needs updating
* I am not really in love with the naming, took me a moment what FocusTop etc means, but I do not have a better idea so its probably fine to leave as is

PR #1031, opened 2026-06-17
Feat/scroll focus

trumpeter122:
## Description

I am sorry that I accidentally closed the original PR #1030. Here is an improved version of it, revised based on you suggestions.

- The keybinds are now under navigation keybinds.
- The scrolling now respects scrolloff.
- The naming has been changed to "scrolling relatively to focused entry". The rationale of this naming is that, such scrolling behavior is defined in the [Vim documentation](https://vimdoc.sourceforge.net/htmldoc/scroll.html) as "Scrolling relative to cursor". Since here we don't have a cursor, I think "focused entry" would be a better option. 

As for updating the documentation, I am a bit confused, because the docs seem to reside in another repo. Should I make another PR there?

### Checklist

- [x] All tests passed


mierak:
Thanks.

> I am sorry that I accidentally closed the original PR

No worries.

> The naming has been changed to "scrolling relatively to focused entry".

Yeah this sounds better, thanks.

> As for updating the documentation, I am a bit confused, because the docs seem to reside in another repo. Should I make another PR there?

Yup, please do. Basically just copying the description you already have in the PR to the kebyinds table [here](https://github.com/rmpc-org/rmpc-org.github.io/blob/master/src/content/docs/configuration/keybinds.mdx) should be enough.

This is good to go when the docs are done, thanks!

trumpeter122:
Thanks for the help. I submitted a [corresponding PR](https://github.com/rmpc-org/rmpc-org.github.io/pull/68) to the docs.

mierak:
Thanks!

PR #1041, opened 2026-07-13
feat: add home manager module with nix-native ron config generator

apothecary103:
## Description
This PR adds a `homeManagerModules.rmpc` flake output that extends the existing Home Manager `programs.rmpc` module by introducing a native Nix-to-RON serializer (`nix/ron.nix`), it removes the need to write raw configuration string blobs. The module introduces new options for `settings`, `themes`, and `ronExtensions`. Because the module isn't bound by a strict upstream schema, any configuration options added to rmpc in the future will work out of the box without requiring updates here.

Standard Nix primitive types (attrsets, lists, strings, booleans, and nulls) map directly to their RON equivalents. However, because plain Nix cannot natively express RON enums or named structs, the module automatically injects an `rmpcLib` argument containing the following helper functions:

| Helper | RON |
| :--- | :--- |
| `variant "Auto"` | `Auto` |
| `enum "Text" [" - "]` | `Text(" - ")` |
| `struct "Split" { size = "50%"; }` | `Split(size: "50%")` |
| `pane "Queue"` | `Pane(Queue)` |
| `tab "Search" "Search"` | `(name: "Search", pane: Pane(Search))` |
| `prop (variant "Artist")` | `Property(Artist)` |
| `text " - "` | `Text(" - ")` |

Helpers can also be nested, e.g., `prop (enum "Status" [(variant "State")])` renders as `Property(Status(State))`

For reference as well, here's my current config that uses this module:
```nix
{ config, rmpcLib, ... }:

let
  inherit (rmpcLib) variant enum struct pane prop text tab;
  bold = fg: { inherit fg; modifiers = "Bold"; };

  # catppuccin macchiato
  mauve = "#c6a0f6";
  lavender = "#b7bdf8";
  sapphire = "#7dc4e4";
  yellow = "#eed49f";
  fgText = "#cad3f5";
  base = "#24273a";
  mantle = "#1e2030";
  overlay0 = "#6e738d";
in
{
  programs.rmpc = {
    enable = true;

    settings = {
      address = "127.0.0.1:6600";
      theme = "catppuccin-macchiato";
      lyrics_dir = "${config.home.homeDirectory}/.lyrics";

      album_art = {
        method = variant "Auto";
        max_size_px = { width = 0; height = 0; };
        vertical_align = variant "Top";
      };

      tabs = [
        {
          name = "Queue";
          pane = struct "Split" {
            direction = variant "Horizontal";
            panes = [
              { size = "100%"; pane = pane "Queue"; }
              {
                size = "42";
                pane = struct "Split" {
                  direction = variant "Vertical";
                  panes = [
                    { size = "20"; borders = "TOP | BOTTOM"; pane = pane "AlbumArt"; }
                    { size = "100%"; borders = "NONE"; pane = pane "Lyrics"; }
                  ];
                };
              }
            ];
          };
        }
        (tab "Directories" "Directories")
        (tab "Artists" "Artists")
        (tab "Album Artists" "AlbumArtists")
        (tab "Albums" "Albums")
        (tab "Playlists" "Playlists")
        (tab "Search" "Search")
      ];
    };

    themes.catppuccin-macchiato = {
      default_album_art_path = null;
      symbols = {
        song = "🎵";
        dir = "📁";
        playlist = "🎼";
        marker = "";
      };

      layout = struct "Split" {
        direction = variant "Vertical";
        panes = [
          { pane = pane "Header"; size = "1"; }
          { pane = pane "TabContent"; size = "100%"; }
          { pane = pane "ProgressBar"; size = "1"; }
        ];
      };

      progress_bar = {
        symbols = [ "" "" "⭘" " " " " ];
        track_style = { bg = mantle; };
        elapsed_style = { fg = mauve; bg = mantle; };
        thumb_style = { fg = mauve; bg = mantle; };
      };

      scrollbar = {
        symbols = [ "│" "█" "▲" "▼" ];
        track_style = { };
        ends_style = { };
        thumb_style = { fg = lavender; };
      };

      browser_column_widths = [ 20 38 42 ];
      text_color = fgText;
      background_color = base;
      header_background_color = mantle;
      modal_background_color = null;
      modal_backdrop = false;

      tab_bar = {
        active_style = { fg = "black"; bg = mauve; modifiers = "Bold"; };
        inactive_style = { };
      };
      borders_style = { fg = overlay0; };
      highlighted_item_style = { fg = mauve; bg = null; modifiers = "Bold"; };
      current_item_style = { fg = "black"; bg = lavender; modifiers = "Bold"; };
      highlight_border_style = { fg = lavender; };

      song_table_format = [
        {
          prop = {
            kind = prop (variant "Artist");
            style = { fg = lavender; };
            default = { kind = text "Unknown"; };
          };
          width = "50%";
          alignment = variant "Right";
        }
        {
          prop = {
            kind = text "-";
            style = { fg = lavender; };
            default = { kind = text "Unknown"; };
          };
          width = "1";
          alignment = variant "Center";
        }
        {
          prop = {
            kind = prop (variant "Title");
            style = { fg = sapphire; };
            default = { kind = text "Unknown"; };
          };
          width = "50%";
        }
      ];

      lyrics = {
        timestamp = false;
        alignment = variant "Center";
      };

      header.rows = [
        {
          left = [
            { kind = text "["; style = bold lavender; }
            { kind = prop (enum "Status" [ (variant "State") ]); style = bold lavender; }
            { kind = text "]"; style = bold lavender; }
          ];
          center = [
            {
              kind = prop (enum "Song" [ (variant "Artist") ]);
              style = bold yellow;
              default = { kind = text "Unknown"; style = bold yellow; };
            }
            { kind = text " - "; }
            {
              kind = prop (enum "Song" [ (variant "Title") ]);
              style = bold sapphire;
              default = { kind = text "No Song"; style = bold sapphire; };
            }
          ];
          right = [
            { kind = text "Vol: "; style = bold lavender; }
            { kind = prop (enum "Status" [ (variant "Volume") ]); style = bold lavender; }
            { kind = text "% "; style = bold lavender; }
          ];
        }
      ];
    };
  };
}
```

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Home Manager module (`homeManagerModules.rmpc` flake output) with structured
+  `programs.rmpc.settings` and `programs.rmpc.themes` options written in plain
+  Nix and rendered to RON at build time.
 - Added a new vertical volume slider pane.
 - **Breaking** `ExternalCommand` can now have arguments supplied at runtime. This will break your
   existing keybinds if they contained either `{` or `}`. You will now need to escape these by
--- a/flake.nix
+++ b/flake.nix
@@ -82,5 +82,16 @@
 
         devShells.default = (mkScope pkgs).shell;
       }
-    );
+    )
+    // {
+      # Structured `programs.rmpc.settings`/`themes` options rendered to RON,
+      # extending the module that ships with Home Manager.
+      homeManagerModules = {
+        rmpc = import ./nix/home-manager.nix;
+        default = inputs.self.homeManagerModules.rmpc;
+      };
+
+      # The Nix -> RON serializer on its own, for use outside Home Manager.
+      lib.ron = import ./nix/ron.nix {inherit (inputs.nixpkgs) lib;};
+    };
 }
--- a/nix/home-manager.nix
+++ b/nix/home-manager.nix
@@ -1,0 +1,130 @@
+# Extends the `programs.rmpc` module that ships with Home Manager
+# (enable/package/config) with structured `settings` and `themes` options
+# rendered to RON. The rendered settings flow through the upstream `config`
+# option, which is `types.lines` and concatenates all definitions — so set
+# either `settings` or `config`, never both.
+{
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.programs.rmpc;
+  ron = import ./ron.nix {inherit lib;};
+
+  # Shorthands for shapes that appear in almost every rmpc config, exposed
+  # together with the ron.nix constructors as the `rmpcLib` module argument.
+  rmpcLib =
+    ron
+    // {
+      # pane "Queue" -> Pane(Queue)
+      pane = name: ron.enum "Pane" [(ron.variant name)];
+
+      # tab "Albums" "Albums" -> (name: "Albums", pane: Pane(Albums))
+      tab = name: pane: {
+        inherit name;
+        pane = rmpcLib.pane pane;
+      };
+
+      # prop (variant "Artist") -> Property(Artist)
+      prop = kind: ron.enum "Property" [kind];
+
+      # text " - " -> Text(" - ")
+      text = s: ron.enum "Text" [s];
+    };
+
+  # Tagged values (variant/enum/struct/raw) are attrsets, admitted by the
+  # attrsOf branch.
+  ronValue = let
+    t = lib.types;
+    valueType =
+      t.nullOr (t.oneOf [
+        t.bool
+        t.int
+        t.float
+        t.str
+        (t.attrsOf valueType)
+        (t.listOf valueType)
+      ])
+      // {
+        description = "RON value";
+      };
+  in
+    valueType;
+
+  renderDocument = value:
+    lib.concatMapStrings (extension: "#![enable(${extension})]\n") cfg.ronExtensions
+    + ron.toRON value
+    + "\n";
+in {
+  options.programs.rmpc = {
+    settings = lib.mkOption {
+      type = lib.types.attrsOf ronValue;
+      default = {};
+      example = lib.literalExpression ''
+        {
+          address = "127.0.0.1:6600";
+          theme = "mytheme";
+          album_art.method = rmpcLib.variant "Kitty";
+          tabs = [
+            (rmpcLib.tab "Queue" "Queue")
+            (rmpcLib.tab "Search" "Search")
+          ];
+        }
+      '';
+      description = ''
+        rmpc configuration, rendered to RON and written to
+        {file}`$XDG_CONFIG_HOME/rmpc/config.ron`. RON constructs with no
+        native Nix representation are built with the helpers provided
+        through the `rmpcLib` module argument. Mutually exclusive with the
+        plain `config` option.
+      '';
+    };
+
+    themes = lib.mkOption {
+      type = lib.types.attrsOf (lib.types.attrsOf ronValue);
+      default = {};
+      example = lib.literalExpression ''
+        {
+          mytheme = {
+            background_color = "#1e1e2e";
+            tab_bar.active_style = { fg = "black"; bg = "#cba6f7"; modifiers = "Bold"; };
+          };
+        }
+      '';
+      description = ''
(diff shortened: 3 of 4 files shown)


mierak:
@Kladki help please :smile: 

If I understand correctly, these are definitions to be able to configure rmpc via .nix?

Kladki:
> @Kladki help please 😄

Hi! 

> If I understand correctly, these are definitions to be able to configure rmpc via .nix?

Yes, but I'm not really too sure what the point of adding a new module here is instead of just using [the one already in home-manager](https://github.com/nix-community/home-manager/blob/master/modules/programs/rmpc.nix) and just overwriting the package, and maybe adding this ron converter to it. @apothecary103?

apothecary103:
> > @Kladki help please 😄
> 
> Hi!
> 
> > If I understand correctly, these are definitions to be able to configure rmpc via .nix?
> 
> Yes, but I'm not really too sure what the point of adding a new module here is instead of just using [the one already in home-manager](https://github.com/nix-community/home-manager/blob/master/modules/programs/rmpc.nix) and just overwriting the package, and maybe adding this ron converter to it. @apothecary103?

Hey there! Thank you for taking a look at the PR!

I did consider submitting this directly to Home Manager. The main issue is the Nix-to-RON serializer, Home Manager usually relies on standard, built-in converters (like JSON or TOML), but because RON requires a custom serializer to handle Rust's enums and structs, getting it merged into Home Manager will likely be quite difficult. They are known to be very slow to review PRs that introduce custom serialization logic because they don't want the maintenance burden.

That said, there's a lot of value in being able to configure rmpc via native Nix. For instance, you can define a colour palette once and pass those variables directly into the config (you can see how that looks in the snippet of my config I attached), trying to manage string interpolation inside a massive RON string becomes a mess pretty quickly. It also ensures that the config is actually valid. If a user misses a comma or a bracket in a raw RON string, Nix won't know at compile time, it will just write the file, and the user will only see an error when they try to launch rmpc. By using native Nix, the serializer guarantees at compile time that everything will work (if there is an issue, Nix will just throw an error and refuse to build). I'm sure there are plenty of other cases where having a native Nix config might be preferred for modularity or programmability.

Kladki:
> I did consider submitting this directly to Home Manager. The main issue is the Nix-to-RON serializer, Home Manager usually relies on standard, built-in converters (like JSON or TOML), but because RON requires a custom serializer to handle Rust's enums and structs, getting it merged into Home Manager will likely be quite difficult. They are known to be very slow to review PRs that introduce custom serialization logic because they don't want the maintenance burden.

I do think it's at least worth trying to get into home-manager, since at least from what I've seen, no one has attempted to do so. Other modules would find it helpful too, after all. Considering that there already are [multiple generators in home-manager](https://github.com/nix-community/home-manager/blob/master/modules/lib/generators.nix), I don't think it should be too difficult, especially since this one for ron is not too long, under 100 lines.

Also, I don't think it's any better to have the maintenance burden split among all repos that have to use RON generators, especially when here the owner doesn't know any nix, it can be really intimidating to have large amounts of code you don't understand in your repo, especially if the person who introduced it is no longer reachable.

> That said, there's a lot of value in being able to configure rmpc via native Nix. [...]

I know the value of having configuration in Nix, I just don't think here is the place for it, instead it should be the home-manager repo.

apothecary103:
> I do think it's at least worth trying to get into home-manager, since at least from what I've seen, no one has attempted to do so. Other modules would find it helpful too, after all. Considering that there already are [multiple generators in home-manager](https://github.com/nix-community/home-manager/blob/master/modules/lib/generators.nix), I don't think it should be too difficult, especially since this one for ron is not too long, under 100 lines.
> 
> Also, I don't think it's any better to have the maintenance burden split among all repos that have to use RON generators, especially when here the owner doesn't know any nix, it can be really intimidating to have large amounts of code you don't understand in your repo, especially if the person who introduced it is no longer reachable.

Very well, I agree that it would be a lot better for the whole ecosystem to have the RON serializer upstream.
I shall take your advice and try to put a PR together over there for the RON generator and see how the maintainers respond to it. Thanks for the feedback!

mierak:
Thank you guys, this can be closed then? Or is there something to add here still when the generator gets merged into home manager?

apothecary103:
I’m closing this for now. I’ll try to upstream it to Home Manager once the generator gets merged.

PR #1049, opened 2026-07-21
fix(browser): render album titles in root_tag shorthand panes

Crimvel:
The `Browser(root_tag: ...)` config shorthand expands its second (album) level with `format: vec![]`, so every album group gets an empty display name and renders as "Untitled".

This gives the album level a real format that renders the `Album` property (with a `<no album>` fallback), matching the built-in Artists pane's album level.

Includes a regression test.


mierak:
such a dumb bug, thanks!

PR #1050, opened 2026-07-21
fix(queue): keep marks on their songs across queue changes

Crimvel:
Queue marks were stored as bare indices into the items vector and re-applied verbatim after each `QueueChanged` event. Removing the first marked songs slid the rest up, so the stale indices landed on whatever songs shifted into those slots — marks appeared to "move" to the next block instead of clearing.

On `QueueChanged`, marks are now translated through stable song ids: remapped to the new indices of the same songs, dropping any whose songs are gone. The remap is factored into `remap_marks_by_id` with unit tests covering removal, reordering, and full drop.

### DIFF
--- a/rmpc/src/ui/panes/queue.rs
+++ b/rmpc/src/ui/panes/queue.rs
@@ -537,7 +537,7 @@ impl Pane for QueuePane {
                 self.queue.unmark_all();
             }
             UiEvent::QueueChanged => {
-                self.queue.items.clone_from(&ctx.queue);
+                remap_marks_by_id(&mut self.queue, &ctx.queue);
             }
             UiEvent::SongChanged => {
                 if let Some(idx) = ctx.current_song_index()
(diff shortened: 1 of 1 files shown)


mierak:
This makes the items weirdly jump around for me: Have a very large queue (10k in my testing) to induce a bit of lag, select ~5 songs that are right next to each other, press and hold for a bit MoveDown, then MoveUp, then MoveDown, ...
I end up with 5 marked songs that are the correct marked ones but they are scattered about. Anything we can do about it?

This functionality has been giving me issues over the years, I am unsure how to handle this correctly because of the idle mechanism of MPD which is slower than user input usually.

Crimvel:
I will look into it. Maybe I find a better solution.

mierak:
Feel free to reach out if you need help with something!

Crimvel:
@mierak So far I couldn't find any new approaches. I will close this PR maybe if I find a way I will resubmit.

mierak:
Alright, thanks!

PR #1055, opened 2026-08-02
fix: rmpcd last fm plugin - convert song duration to seconds for now playing

frumpy4:
The song duration is in milliseconds but lastfm's now playing endpoint expects seconds. It sets your now playing status for the duration given in the request, so it was being set for 1000x longer than it should've been

I made sure it doesn't need to be an integer, for example sending the duration "30.123" sets it for 30 seconds


mierak:
Thanks!

PR #1065, opened 2026-08-18
Add a `Hash` property transform to color content from a palette

### LINKED ISSUE #839: Style by hash of content
### Description

This feature request is a bit out there... But I think it'd be neat.

I'd like to be able to style things (their colour in particular) based on their content. So for example different years appear in different colours, but the same year appears in the same colour everywhere.  I might want to use this on artists, years, albums in a list view for example to make them contrast more.

### Proposed solution

- Start with a cheap hash of the content which will be printed (or perhaps optionally of a separate piece of configured content).
- Represent that hash as an integer.
- Take the configured list of allowed colours, and choose an index based on `hash_int % len(allowed_colours)`.
- Use that colour in the style.

ChrisJr404:
Closes #839.

### What

Adds a new `Hash` property transform. It renders its `content` and applies a foreground color chosen from a configured `colors` palette based on a stable hash of the rendered text. The same value always maps to the same color while different values spread across the palette, which makes it easy to tell artists, years or albums apart at a glance in a list view.

```ron
Transform(Hash(
    content: Property(Artist),
    colors: ["red", "green", "blue", "magenta", "cyan"],
))
```

### How

It mirrors the existing `Truncate`/`Replace` transforms:

- a new `Hash` variant on both `TransformFile` (config) and `Transform` (runtime); the palette strings are parsed into `ratatui` colors once, during config conversion, in both `properties.rs` and `queue_table.rs`;
- render hooks in the string / span / line paths (`panes/mod.rs`, `song_ext.rs`). The string path returns the content unchanged; the span and line paths override the foreground with the hashed color and leave the rest of the style intact.

The color is selected with a small FNV-1a hash so the mapping is stable across runs and platforms. An empty palette is a no-op (the content stays unstyled), so it degrades gracefully.

### Tests

- unit tests for the selection helper: determinism, palette membership, empty palette, single-color palette, and spread across the palette;
- `as_span` tests for content pass-through, deterministic coloring from the palette, and the empty-palette no-op.

`cargo test`, `cargo fmt` (nightly) and `cargo clippy` are all clean.

### DIFF
--- a/rmpc/src/config/theme/properties.rs
+++ b/rmpc/src/config/theme/properties.rs
@@ -3,14 +3,39 @@ use std::collections::BTreeMap;
 use anyhow::{Result, bail};
 use bon::Builder;
 use itertools::Itertools;
-use ratatui::style::Style;
+use ratatui::style::{Color, Style};
 use rmpc_mpd::filter::Tag;
 use serde::{Deserialize, Serialize};
 use serde_with::skip_serializing_none;
 use strum::Display;
 
-use super::style::ToConfigOr;
+use super::style::{StringColor, ToConfigOr};
 use crate::config::{defaults, theme::StyleFile};
+
+/// Picks a color from `colors` deterministically based on the content string.
+///
+/// The same content always maps to the same color, while different content is
+/// spread across the palette. Returns `None` when `colors` is empty so callers
+/// can leave the content unstyled.
+pub(crate) fn hashed_color(content: &str, colors: &[Color]) -> Option<Color> {
+    if colors.is_empty() {
+        return None;
+    }
+
+    // FNV-1a, a small deterministic hash so the mapping is stable across runs.
+    let mut hash: u64 = 0xcbf2_9ce4_8422_2325;
+    for byte in content.as_bytes() {
+        hash ^= u64::from(*byte);
+        hash = hash.wrapping_mul(0x0000_0100_0000_01b3);
+    }
+
+    let idx = (hash % colors.len() as u64) as usize;
+    Some(colors[idx])
+}
+
+pub(super) fn parse_colors(colors: Vec<String>) -> Result<Vec<Color>> {
+    colors.into_iter().filter_map(|color| StringColor(Some(color)).to_color().transpose()).collect()
+}
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub enum SongPropertyFile {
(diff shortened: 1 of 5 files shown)


rmpc/src/config/theme/properties.rs
```diff
@@ -3,15 +3,40 @@ use std::collections::BTreeMap;
...
-use super::style::ToConfigOr;
+use super::style::{StringColor, ToConfigOr};
 use crate::config::{defaults, theme::StyleFile};
 
+/// Picks a color from `colors` deterministically based on the content string.
+///
+/// The same content always maps to the same color, while different content is
+/// spread across the palette. Returns `None` when `colors` is empty so callers
+/// can leave the content unstyled.
+pub(crate) fn hashed_color(content: &str, colors: &[Color]) -> Option<Color> {
```
mierak, line 20: This should live in the ui module not the config since its used during the application's lifecycle and not on startup.

rmpc/src/config/theme/properties.rs
```diff
@@ -294,12 +319,17 @@ pub enum TransformFile<T: Clone> {
         content: Box<PropertyFile<T>>,
         replacements: Vec<ReplacementFile<T>>,
     },
+    Hash {
```
mierak, line 322: The name Hash does not really describe the behavior at all. Something like Colorize/Palette/PaletteColor or something similar would be better.

rmpc/src/config/theme/properties.rs
```diff
@@ -294,12 +319,17 @@ pub enum TransformFile<T: Clone> {
...
+        content: Box<PropertyFile<T>>,
+        colors: Vec<String>,
```
mierak, line 324: This should be a whole Style instead of a simple fg color. Gives more freedom to users and lets them avoid contrast issues.

mierak (commented):
Thanks! 

Documentation is needed before this can be merged too.

ChrisJr404:
Rebased onto master to clear the conflict. Docs live in the separate site repo now, so I added the Hash transform to the Transforms table there in rmpc-org/rmpc-org.github.io#75. Thanks!

mierak:
What about the other points?

mierak:
Hi, are you still going to finish this?

ChrisJr404:
Yes, still on it, sorry for the delay. I agree with all three points: I will move it into the ui module since it runs during the application lifecycle rather than at startup, rename it to something that describes the behavior (Palette or Colorize rather than Hash), and take a full Style instead of just a foreground color so users can avoid contrast issues. I will push the revision.

PR #1067, opened 2026-08-19
feat: add `current` option to Save action to save currently playing song

### LINKED ISSUE #1066: feat: add current playing song to a playlist from any pane
### Description

**Is your feature request related to a problem? Please describe.**
Right now, adding a song to a playlist via the `Save` keybind always operates on whatever song is under the cursor (or marked/selected) in the currently focused pane. There's no way to save the *currently playing* song to a playlist unless you first navigate to the Queue tab and move the cursor onto it (e.g. via `JumpToCurrent`).

This is inconvenient when browsing, say, the Artists or Search tab while a song plays — you have to switch tabs, jump the cursor to the current song, and only then save it, instead of doing it in place.

**Describe the solution you'd like**
Add a `current: bool` option to the `Save` action (mirroring how `Rate` already supports `current: true`/`false`). When `current: true`, the action ignores cursor position/selection entirely and instead saves whatever song MPD reports as currently playing (`ctx.current_song()`), regardless of which pane/tab is focused.

Example config:
```ron
"<C-s>S": Save(kind: Modal(all: false, duplicates_strategy: Ask), current: true),
```

This lets users bind a single global-ish keybind to "add currently playing song to playlist" that works everywhere, the same way `Rate(current: true)` already lets you rate the currently playing song from anywhere.

**Describe alternatives you've considered**
- Binding `JumpToCurrent` + `Save` as two separate keypresses — works, but requires being on the Queue tab and two actions instead of one.
- Using `rmpc remote keybind` to chain `JumpToCurrent` then `Save` via an external script — works but is a workaround rather than a native feature, and doesn't survive to `Save(kind: Playlist(name: ...))` for direct single-key save-to-a-named-playlist use cases.

**Additional context**
I already have a working patch for this against master (adds `current: bool` to `CommonActionFile::Save`/`CommonAction::Save`, handles it in `ui/mod.rs` with the panes abandoning the event via `event.abandon()` when `current: true`, same pattern already used for `Rate`). Happy to open a PR if this approach sounds reasonable — let me know if you'd prefer a different design (e.g. a dedicated new action instead of extending `Save`).

### Proposed solution

_No response_

nestthub:
Closes #1066 

Adds a `current: bool` field to the `Save` keybind action, mirroring the existing `Rate(current: bool)` option. When `current: true`, the action ignores the cursor position and selection in the focused pane and instead saves the song currently playing (as reported by `ctx.current_song()`) to a playlist, either via the picker modal (`SaveKind::Modal`) or directly to a named playlist (`SaveKind::Playlist`).

This allows binding a single keybind to "add currently playing song to playlist" that works from any pane/tab, without first having to jump to the Queue tab and move the cursor onto the playing song.

Implementation follows the same pattern already used for `Rate(current: true)`: each pane's `Save` handler now matches `current: true` first and calls `event.abandon()`, letting the event bubble up to the shared handler in `ui/mod.rs`, which resolves the currently playing song and opens the save modal or saves directly.

Changes:
- config/keys/actions.rs: add `current: bool` to `CommonActionFile::Save` and `CommonAction::Save`, update `TryFrom` conversion and `to_description` text.
- config/keys/mod.rs: update default keybinds for the new field and add a new default binding for saving the current song.
- ui/mod.rs: handle `CommonAction::Save { current: true, .. }` globally, resolving `ctx.current_song()` and reusing `create_save_modal` / `add_to_playlist_or_show_modal`.
- ui/browser.rs, ui/panes/queue.rs, ui/panes/search/mod.rs: abandon the `Save` event when `current: true` so it bubbles up to the shared handler; explicit `current: false` added to existing arms.

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Added `current` option to `Save()` keybind to save the currently playing song
 - Added a new vertical volume slider pane.
 - **Breaking** `ExternalCommand` can now have arguments supplied at runtime. This will break your
   existing keybinds if they contained either `{` or `}`. You will now need to escape these by
--- a/assets/example_config.ron
+++ b/assets/example_config.ron
@@ -119,6 +119,7 @@
             "oi":         ShowInfo,
             "<C-x>":      ContextMenu(),
             "<C-s>s":     Save(kind: Modal(all: false, duplicates_strategy: Ask)),
+            "<C-s>c": Save(kind: Modal(all: false, duplicates_strategy: Ask), current: true),
             "<C-s>a":     Save(kind: Modal(all: true, duplicates_strategy: Ask)),
             "r":          Rate(),
             "Y":          CopyToClipboard(kind: Modal([
--- a/assets/example_config.ron
+++ b/assets/example_config.ron
@@ -119,7 +119,7 @@
             "oi":         ShowInfo,
             "<C-x>":      ContextMenu(),
             "<C-s>s":     Save(kind: Modal(all: false, duplicates_strategy: Ask)),
-            "<C-s>c": Save(kind: Modal(all: false, duplicates_strategy: Ask), current: true),
+            "<C-s>c":     Save(kind: Modal(all: false, duplicates_strategy: Ask), current: true),
             "<C-s>a":     Save(kind: Modal(all: true, duplicates_strategy: Ask)),
             "r":          Rate(),
             "Y":          CopyToClipboard(kind: Modal([
--- a/rmpc/src/config/keys/actions.rs
+++ b/rmpc/src/config/keys/actions.rs
@@ -990,6 +990,8 @@ pub enum CommonActionFile {
     Save {
         #[serde(default)]
         kind: SaveKind,
+        #[serde(default)]
+        current: bool,
     },
     DeleteFromPlaylist {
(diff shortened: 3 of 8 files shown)


mierak:
Thanks.

The code looks mostly ok. Needs a reformat, PR to the documentation, changelog update and fix tests before this can get merged.

nestthub:
Hello. I rushed the PR a bit. I went through everything again and completed points 1–3 and 5.

I have a question regarding point 4. As I understand it, I need to update the documentation in [https://github.com/rmpc-org/rmpc-org.github.io](https://github.com/rmpc-org/rmpc-org.github.io). Am I correct that I should update the documentation in the `src/content/docs/configuration/` directory on the `master` branch, rather than the documentation for version `v0.11.0`?

Also, I’m not entirely sure about the order of the changes. Should I update the documentation first and then make the corresponding changes in the code, or should the code changes come first and the documentation be updated afterwards?

 Thanks in advance for the clarification.

mierak:
> Am I correct that I should update the documentation in the `src/content/docs/configuration`

correct!



> Also, I’m not entirely sure about the order of the changes.

Just create two PRs ar the same time, one in each repo. The order doesnt really matter.

nestthub:
Done 🙌🏽
[PR #73](https://github.com/rmpc-org/rmpc-org.github.io/pull/73)

mierak:
Thanks!

PR #1072, opened 2026-08-20
feat: add output status properties and improve the outputs modal

Strykar:
### Description

MPD's `output` idle event was received and discarded
(`IdleEvent::Output => {}`) and no theme property could display which
outputs are enabled. This PR stores the current partition's outputs on
`Ctx` (fetched at startup, on the `output` idle event and on reconnect)
and adds two status properties.

`ActiveOutput` renders a fixed-width on/off lamp for one named output,
with the usual label and style pairs (the labels default to the
output's name). Composed one per output in a `Group`, the display keeps
a fixed width no matter how many outputs are enabled:

```ron
(kind: Property(Status(ActiveOutput(name: "S/PDIF",
    on_style: (fg: "yellow", modifiers: "Bold"), off_style: (fg: "gray"))))),
(kind: Text(" ")),
(kind: Property(Status(ActiveOutput(name: "Speakers", on_label: "SPKR", off_label: "SPKR",
    on_style: (fg: "yellow", modifiers: "Bold"), off_style: (fg: "gray"))))),
```

ActiveOutput lamps: one per output, fixed width; enabled outputs light up

The always-visible dimmed look is one choice, not a requirement: an
empty `off_label` hides a lamp entirely while its output is disabled
(put the spacing inside `on_label`), so space-constrained layouts show
only what is playing.

`ActiveOutputs` renders the enabled output names joined by a separator:

```ron
(kind: Property(Status(ActiveOutputs(separator: ", "))))
```

Why both: MPD plays to every enabled output at once, so the summary can
grow past a fixed pane and truncate (found in live testing). The lamp
row cannot grow, and per-output lamps cannot be built from the joined
string with `Transform`. The summary stays because it is the only
display that reveals an output the theme did not anticipate.

Behavior details:

- The fetch is a plain `outputs` command: it reflects the partition the
  connection is on and never switches partitions. A switch inside the
  fetch would raise another `output` idle event on the same connection
  and loop, and partition switches need no extra trigger because MPD's
  `Client::SetPartition` flags `IDLE_OUTPUT` on the switching client.
- The summary excludes outputs on other partitions (MPD reports them as
  the `dummy` plugin). A lamp for an output that was moved to another
  partition renders `off_label`; only an unknown name falls through to
  the item's `default:`, same as `Bitrate` and `Crossfade`. An empty
  summary resolves to `default:` too.
- Non-audio outputs (fifo, httpd) count as enabled. Lamps handle this
  naturally (do not author a lamp for them); for the summary a
  `Transform` arm can hide them, and the parsed `plugin` field is
  available if a built-in filter is preferred later.

Scope note: the outputs modal keeps its own fetch-on-open and
refresh-on-event queries, so while it is open an output change fetches
the list twice (once for the modal, once for the shared state). Left
that way deliberately to keep this PR from rewriting existing modal
behavior; a follow-up could render the modal from the shared outputs
state and drop its private copy. Happy to do that here instead if
preferred.

The second commit makes the modal refresh on reconnect; it only
refreshed on the output idle event, so a modal left open across an MPD
restart kept the pre-disconnect outputs.

My use case: one `pipewire` output per physical route pinned by
`target` in mpd.conf. The lamp row shows the active route at a glance,
toggling in the outputs modal switches it live, and it works for remote
MPD where host-side tools cannot answer this.

### Related issues

None found for output display in the status bar or theme.

### Checklist

- [x] All tests passed
- [x] Code has been checked with clippy (stable, clean)
- [x] [Documentation](https://github.com/rmpc-org/rmpc-org.github.io) has
      been updated (companion PR: rmpc-org/rmpc-org.github.io#76)

### DIFF
--- a/rmpc/src/config/theme/properties.rs
+++ b/rmpc/src/config/theme/properties.rs
@@ -148,6 +148,21 @@ pub enum StatusPropertyFile {
     SampleRate(),
     Bits(),
     Channels(),
+    ActiveOutputs {
+        #[serde(default = "defaults::default_output_separator")]
+        separator: String,
+    },
+    ActiveOutput {
+        name: String,
+        #[serde(default)]
+        on_label: Option<String>,
+        #[serde(default)]
+        off_label: Option<String>,
+        #[serde(default)]
+        on_style: Option<StyleFile>,
+        #[serde(default)]
+        off_style: Option<StyleFile>,
+    },
 }
 
 #[derive(Debug, Clone, Display, Hash, Eq, PartialEq)]
@@ -209,6 +224,16 @@ pub enum StatusProperty {
     SampleRate(),
     Bits(),
     Channels(),
+    ActiveOutputs {
+        separator: String,
+    },
+    ActiveOutput {
+        name: String,
+        on_label: String,
+        off_label: String,
+        on_style: Option<Style>,
+        off_style: Option<Style>,
+    },
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -521,6 +546,22 @@ impl TryFrom<StatusPropertyFile> for StatusProperty {
             StatusPropertyFile::SampleRate() => StatusProperty::SampleRate(),
             StatusPropertyFile::Bits() => StatusProperty::Bits(),
             StatusPropertyFile::Channels() => StatusProperty::Channels(),
+            StatusPropertyFile::ActiveOutputs { separator } => {
+                StatusProperty::ActiveOutputs { separator }
+            }
+            StatusPropertyFile::ActiveOutput { name, on_label, off_label, on_style, off_style } => {
+                StatusProperty::ActiveOutput {
+                    on_label: on_label.unwrap_or_else(|| name.clone()),
+                    off_label: off_label.unwrap_or_else(|| name.clone()),
+                    name,
+                    on_style: on_style
+                        .map(|s| -> Result<_> { s.to_config_or(None, None) })
+                        .transpose()?,
+                    off_style: off_style
+                        .map(|s| -> Result<_> { s.to_config_or(None, None) })
+                        .transpose()?,
+                }
+            }
         })
     }
 }
--- a/rmpc/src/core/event_loop.rs
+++ b/rmpc/src/core/event_loop.rs
@@ -26,6 +26,7 @@ use crate::{
         mpd_client_ext::MpdClientExt,
         mpd_query::{
             EXTERNAL_COMMAND,
+            GLOBAL_OUTPUTS_UPDATE,
             GLOBAL_QUEUE_UPDATE,
             GLOBAL_STATUS_UPDATE,
             GLOBAL_STICKERS_UPDATE,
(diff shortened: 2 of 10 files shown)


rmpc/src/ui/modals/outputs.rs
```diff
@@ -102,7 +102,8 @@ impl Modal for OutputsModal {
             .border_set(border::ROUNDED)
             .border_style(ctx.config.as_border_style())
             .title_alignment(ratatui::prelude::Alignment::Center)
-            .title("Outputs");
+            .title("Outputs")
+            .title_bottom(" Enter toggles; all enabled outputs play at once ");
```
mierak: This is simply not true, the bindings are fully dependent on user config, enter is just the default.
I say remove the bottom title completely. Its not consistent with the rest of rmpc anyway.

rmpc/src/config/theme/properties.rs
```diff
@@ -148,6 +148,21 @@ pub enum StatusPropertyFile {
...
+    ActiveOutputs {
+        #[serde(default = "defaults::default_output_separator")]
+        separator: String,
+    },
+    ActiveOutput {
+        name: String,
+        #[serde(default = "defaults::default_on_label")]
+        on_label: String,
+        #[serde(default = "defaults::default_off_label")]
+        off_label: String,
```
mierak:
```suggestion
        #[serde(default)]
        on_label: Option<String>,
        #[serde(default)]
        off_label: Option<String>,
```
And default them to the `name` during conversion to the runtime enum(`StatusProperty`). Without this, putting multiple `ActiveOutput` next to each other with different names and without specifying custom labels results in eg. `On On On` which is not really desirable.

rmpc/src/core/event_loop.rs
```diff
@@ -931,7 +941,15 @@ fn handle_idle_event(event: IdleEvent, ctx: &Ctx) {
                 })
             });
         }
-        IdleEvent::Output => {}
+        IdleEvent::Output => {
+            // A plain outputs fetch: it reflects whatever partition the
```
mierak: remove comment

rmpc/src/ui/panes/mod.rs
```diff
@@ -718,6 +718,34 @@ impl Property<PropertyKind> {
...
+                        .map(|output| output.name.as_str())
+                        .join(separator);
+                    if outputs.is_empty() {
+                        self.default_as_span(song, ctx, tag_separator, strategy)
+                    } else {
+                        Some(Either::Left(Span::styled(outputs, style)))
+                    }
+                }
+                StatusProperty::ActiveOutput { name, on_label, off_label, on_style, off_style } => {
+                    // An output that exists but sits on another partition is
```
mierak: remove comment

rmpc/src/ui/modals/outputs.rs
Strykar: Right, dropped that commit entirely.

rmpc/src/config/theme/properties.rs
Strykar: Done. The labels fall back to the output's name in the `TryFrom` conversion, so a bare `ActiveOutput(name: "S/PDIF")` now renders the name in both states. An explicitly empty `off_label` still renders nothing while the output is disabled; both paths are covered by tests that go through the conversion.

mierak:
Thanks. I left a review in the docs PR, this should be good to go after that is taken care of.

PR #1076, opened 2026-08-24
feat(directories): blacklist folders from the browser pane

Crimvel:
Adds a `browser_hidden_dirs` config option: a list of directory names to hide from the Directories browser pane.

Names are matched against each directory's leaf name at any depth, so e.g. `browser_hidden_dirs: [".Trash-1000"]` keeps trash folders out of the browser no matter where they appear in the library. Only filesystem directories are filtered — songs and playlists sharing a hidden name are untouched.

- New `Vec<String>` option in `ConfigFile`, defaults to empty (no behavior change unless set)
- Filtering applied at root listing, database refresh, and nested `fetch_data`
- Example config updated, changelog entry added, unit test for the leaf-name matching

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `directories_hidden_dirs` config option to hide directories with the given names from the
+  `Directories` pane
 - Added `current` option to `Save()` keybind to save the currently playing song
 - Added a new vertical volume slider pane.
 - **Breaking** `ExternalCommand` can now have arguments supplied at runtime. This will break your
--- a/assets/example_config.ron
+++ b/assets/example_config.ron
@@ -26,6 +26,7 @@
     ignore_leading_the: false,
     browser_song_sort: [Disc, Track, Artist, Title],
     directories_sort: SortFormat(group_by_type: true, reverse: false),
+    directories_hidden_dirs: [],
     auto_open_downloads: true,
     quit_closes_modal: false,
     queue_disable_current_item_style_timeout_ms: None,
--- a/rmpc/src/config/mod.rs
+++ b/rmpc/src/config/mod.rs
@@ -84,6 +84,7 @@ pub struct Config {
     pub browser_song_sort: Arc<SortOptions>,
     pub show_playlists_in_browser: ShowPlaylistsMode,
     pub directories_sort: Arc<SortOptions>,
+    pub directories_hidden_dirs: Arc<Vec<String>>,
     pub cava: Cava,
     pub auto_open_downloads: bool,
     pub extra_yt_dlp_args: Vec<String>,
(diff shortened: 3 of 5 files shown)


mierak:
Hm, is this feature really needed? I cannot imagine a problem this solves other than the presented one (some hidden/system directories in your music dir).

Wouldn't something like `directories_show_hidden_<dirs/entries>` which hides anything starting with `.` be enough?
Not against this per se if its really needed to solve your issue, but maybe this is a better solution?

Also, a note `browser_hidden_dirs` implies it works in every browser pane, not just in the directories, that would have to be changed to `directories_hidden_dirs`.

Crimvel:
directories_hidden_dirs makes more sense I agree. 

Maybe my problem is more unique than I thought, I have a few folders that I don't want to show up in the Dir Pane (.Trash was just the first that annoyed the hell out of me).

assets/example_config.ron
```diff
@@ -26,6 +26,9 @@
     ignore_leading_the: false,
     browser_song_sort: [Disc, Track, Artist, Title],
     directories_sort: SortFormat(group_by_type: true, reverse: false),
+    // Directory names hidden from the Directories browser pane. Matched against
+    // each directory's leaf name at any depth, e.g. [".Trash-1000"].
+    browser_hidden_dirs: [],
```
mierak: The comments are redundant I think. They should be mostly self explanatory and if not, there is always documentation site.

rmpc/src/config/mod.rs
```diff
@@ -156,6 +157,10 @@ pub struct ConfigFile {
     pub browser_song_sort: Vec<SongPropertyFile>,
     pub show_playlists_in_browser: ShowPlaylistsMode,
     pub directories_sort: SortModeFile,
+    /// Directory names hidden from the Directories browser pane. Matched
+    /// against each directory's leaf name at any depth (e.g.
+    /// `.Trash-1000`).
+    pub browser_hidden_dirs: Vec<String>,
```
mierak: same here

assets/example_config.ron
mierak:
```suggestion
    directories_hidden_dirs: [],
```

rmpc/src/ui/dir_or_song.rs
```diff
@@ -84,6 +84,13 @@ impl DirOrSong {
         }
     }
 
+    /// Whether this entry is a (non-playlist) directory whose leaf name appears
+    /// in `hidden`. Used to blacklist folders like `.Trash-1000` from the
+    /// Directories browser pane.
```
mierak: Again, I think its mostly just a noise that will become stale at some point. The code is mostly self explanatory.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `browser_hidden_dirs` config option to hide directories with the given names from the
```
mierak: update after changing the property name

mierak (commented):
> Maybe my problem is more unique than I thought, I have a few folders that I don't want to show up in the Dir Pane (.Trash was just the first that annoyed the hell out of me).

Alright. This is mostly good to go then. Docs PR is needed too.

Crimvel:
Renamed to `directories_hidden_dirs` and dropped the redundant comments as suggested. Docs PR: rmpc-org/rmpc-org.github.io#77

Crimvel:
Thanks @mierak if there is anything else please tell me.

mierak:
Thanks!

PR #1078, opened 2026-08-25
fix: rmpcd - restore channel subscriptions after reconnect

Strykar:
## Description
MPD maintains channel subscriptions per connection. When the connection drops, every subscription made on it is gone. `rmpcd` only subscribes in one loop at `startup`, (`main.rs`, right after `mpd.connect`). It never resubscribes again.

`async_client.rs` recovers the connection and `PluginEvent::Reconnect` is sent to Lua plugins, but subscribe commands are not re-issued.

So after any / every reconnect it results in `rmpcd` connected and appearing healthy, but every plugin channel is unsubscribed.

```sh
RPID=$(systemctl --user show -p MainPID --value rmpcd.service)
PORT=$(ss -tanp | grep "pid=$RPID" | grep -oP '127\.0\.0\.1:\K[0-9]+(?=\s)' | head -1)
sudo ss -K src 127.0.0.1:$PORT dst 127.0.0.1:6600
sleep 5
mpc channels
```

Before the patch, `rmpcd` reconnects and `mpc channels` comes back empty:

```
WARN  Lost MPD connection, attempting reconnect error=Software caused connection abort (os error 103)
INFO  Reconnected to MPD
```

```
$ mpc channels
$ mpc sendmessage rmpcd.lyrics test
MPD error: nobody is subscribed to this channel
```

After the patch all five channels are listed again and `sendmessage` works:

```
WARN  Lost MPD connection, attempting reconnect error=Software caused connection abort (os error 103)
INFO  Reconnected to MPD
INFO  Resubscribed to channel channel="rmpcd.lastfm"
INFO  Resubscribed to channel channel="rmpcd.playcount"
INFO  Resubscribed to channel channel="rmpcd.notify"
INFO  Resubscribed to channel channel="rmpcd.lyrics"
INFO  Resubscribed to channel channel="rmpcd.favorites"
```
cargo +nightly fmt --all --check clean, cargo clippy -p rmpcd --release clean.
Assisted-by Claude Opus 5

### Checklist

- [x] All tests passed

### DIFF
--- a/rmpcd/src/async_client.rs
+++ b/rmpcd/src/async_client.rs
@@ -4,7 +4,6 @@ use std::{
     sync::{
         Arc,
         Mutex as StdMutex,
-        PoisonError,
         atomic::{AtomicU8, Ordering},
     },
     thread,
@@ -52,16 +51,14 @@ struct Shared {
     #[debug(skip)]
     interrupt_stream: StdMutex<Option<MpdStream>>,
     idle_state: AtomicU8,
-    /// Channels this client is subscribed to. MPD forgets subscriptions when
-    /// the connection goes away, so they have to be restored on reconnect.
-    subscriptions: StdMutex<HashSet<String>>,
+    subscribed_channels: StdMutex<HashSet<String>>,
 }
 
 fn resubscribe(client: &mut MpdClient, shared: &Shared) {
     let channels: Vec<String> = shared
-        .subscriptions
+        .subscribed_channels
         .lock()
-        .unwrap_or_else(PoisonError::into_inner)
+        .expect("Failed to lock subscribed channels")
         .iter()
         .cloned()
         .collect();
@@ -218,7 +215,7 @@ impl AsyncClient {
         let shared = Arc::new(Shared {
             interrupt_stream: StdMutex::new(None),
             idle_state: AtomicU8::new(IDLE_STATE_NOT_IDLE),
-            subscriptions: StdMutex::new(HashSet::new()),
+            subscribed_channels: StdMutex::new(HashSet::new()),
         });
 
         let init =
@@ -276,13 +273,13 @@ impl AsyncClient {
     pub async fn subscribe(&self, channel: String) -> Result<(), MpdError> {
         let shared = Arc::clone(&self.shared);
 
-        // Recorded inside the closure so it happens on the worker thread, in the
-        // same step as the command. Doing it after run() returns leaves a window
-        // where a reconnect can snapshot the set before the caller records, and
-        // run()'s timeout can abandon a command that later executes anyway.
         self.run(move |c| {
             c.subscribe(&channel)?;
-            shared.subscriptions.lock().unwrap_or_else(PoisonError::into_inner).insert(channel);
+            shared
+                .subscribed_channels
+                .lock()
+                .expect("Failed to lock subscribed channels")
+                .insert(channel);
             Ok(())
         })
         .await
@@ -293,7 +290,11 @@ impl AsyncClient {
 
         self.run(move |c| {
             c.unsubscribe(&channel)?;
-            shared.subscriptions.lock().unwrap_or_else(PoisonError::into_inner).remove(&channel);
+            shared
+                .subscribed_channels
+                .lock()
+                .expect("Failed to lock subscribed channels")
+                .remove(&channel);
             Ok(())
         })
         .await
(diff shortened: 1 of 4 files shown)


rmpcd/src/async_client.rs
```diff
@@ -249,6 +273,32 @@ impl AsyncClient {
...
+    pub async fn subscribe(&self, channel: String) -> Result<(), MpdError> {
+        let shared = Arc::clone(&self.shared);
+
+        // Recorded inside the closure so it happens on the worker thread, in the
+        // same step as the command. Doing it after run() returns leaves a window
+        // where a reconnect can snapshot the set before the caller records, and
+        // run()'s timeout can abandon a command that later executes anyway.
+        self.run(move |c| {
+            c.subscribe(&channel)?;
+            shared.subscriptions.lock().unwrap_or_else(PoisonError::into_inner).insert(channel);
```
mierak: can this be changed to a lock().expect()? this is something that should never happen and silencing it seems not that great

rmpcd/src/async_client.rs
```diff
@@ -249,6 +273,32 @@ impl AsyncClient {
...
+        })
+        .await
+    }
+
+    pub async fn unsubscribe(&self, channel: String) -> Result<(), MpdError> {
+        let shared = Arc::clone(&self.shared);
+
+        self.run(move |c| {
+            c.unsubscribe(&channel)?;
+            shared.subscriptions.lock().unwrap_or_else(PoisonError::into_inner).remove(&channel);
```
mierak: same here

rmpcd/src/async_client.rs
```diff
@@ -249,6 +273,32 @@ impl AsyncClient {
             .map_err(|err| MpdError::Generic(format!("Worker dropped response: {err}")))?
     }
 
+    pub async fn subscribe(&self, channel: String) -> Result<(), MpdError> {
+        let shared = Arc::clone(&self.shared);
+
+        // Recorded inside the closure so it happens on the worker thread, in the
```
mierak: These comments are not really adding any value, remove them.

rmpcd/src/async_client.rs
```diff
@@ -50,6 +52,26 @@ struct Shared {
     #[debug(skip)]
     interrupt_stream: StdMutex<Option<MpdStream>>,
     idle_state: AtomicU8,
+    /// Channels this client is subscribed to. MPD forgets subscriptions when
+    /// the connection goes away, so they have to be restored on reconnect.
+    subscriptions: StdMutex<HashSet<String>>,
```
mierak:
```suggestion
    subscribed_channels: StdMutex<HashSet<String>>,
```
And remove the comment, its redundant.

rmpcd/src/async_client.rs
```diff
@@ -162,6 +184,7 @@ fn worker_loop(
                                     }
                                 }
                                 info!("Reconnected to MPD");
+                                resubscribe(&mut client, &shared);
```
mierak, line 184: The subscription can fail, for example if the client is already subscribed to the channel already. This should probably be treated as an error when rmpcd fails to resubscribe on reconnect and maybe let the reconnect loop again?

Strykar: Agreed that a silent failure here is wrong, but I'd like to scope it rather than always re-entering the reconnect loop.

On the "already subscribed" case specifically: I don't think `resubscribe` can hit that. A reconnect gives us a fresh connection with no subscriptions on the MPD side, and the source is a `HashSet`, so it can't contain duplicates. The only way to see that error is a plugin's own `reconnect` hook subscribing on top of what we just restored, which is the plugin's call rather than this loop's.

The failures I think we can actually hit split in two:

- Connection-class (`ClientClosed`, or the io-derived variants). The connection went away again mid-restore. Worth retrying, though note it's already self-healing today: the `enter_idle` immediately after fails, and that drops us into the reconnect loop anyway.
- `MpdError::Mpd(MpdFailureResponse`. MPD rejected the command itself, e.g. a channel name it won't accept. That fails identically on every attempt, so re-entering the reconnect loop on it would spin forever on a name that is never going to work.

So rather than treating any resubscribe failure as a reconnect trigger, I'd match on the error and only loop again for the connection-class ones, logging and continuing for `MpdError::Mpd`.

Does this sound right, happy to implement it either ways?

mierak: Alright, lets leave it as is, the subscription failures are logged at the very least os it should be enough for now hopefully.

Strykar: Ok, your call.

rmpcd/src/async_client.rs
Strykar: Fixed in 3 places, thanks for the review!

PR #1080, opened 2026-08-27
fix: rmpcd - same channel on multiple plugins

mierak:
## Description

A bit of continuation on work on correcting the MPD channel subscription that was done by @Strykar in https://github.com/mierak/rmpc/pull/1078.

This essentially refcounts the channel subscriptions since there is nothing stopping multiple plugins from sharing the same channel and before this change, rmpcd would fail to start when the second plugin initializes. The channel is only subscribed to when subscription is asked for the first time. Likewise, it is only unsubscribed from when there are no more subscriptions. Otherwise the sub/unsub are simple noops.

One issue remaining is when a plugin acts badly and repeatedly unsubscribes from a channel that another plugin is using, but that is an ok limitation in my book.

### Checklist

- [x] All tests passed

### DIFF
--- a/rmpcd/src/async_client.rs
+++ b/rmpcd/src/async_client.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::HashSet,
     io::Write,
     sync::{
         Arc,
@@ -21,6 +20,8 @@ use tokio::{
     time::timeout,
 };
 use tracing::{error, info, trace, warn};
+
+use crate::subscribed_channels::SubscribedChannels;
 
 pub type MpdClient = rmpc_mpd::client::Client<'static>;
 pub type MpdError = rmpc_mpd::errors::MpdError;
@@ -51,19 +52,11 @@ struct Shared {
     #[debug(skip)]
     interrupt_stream: StdMutex<Option<MpdStream>>,
     idle_state: AtomicU8,
-    subscribed_channels: StdMutex<HashSet<String>>,
+    subscribed_channels: SubscribedChannels,
 }
 
 fn resubscribe(client: &mut MpdClient, shared: &Shared) {
-    let channels: Vec<String> = shared
-        .subscribed_channels
-        .lock()
-        .expect("Failed to lock subscribed channels")
-        .iter()
-        .cloned()
-        .collect();
-
-    for channel in channels {
+    for channel in shared.subscribed_channels.list() {
         match client.subscribe(&channel) {
             Ok(()) => info!(channel, "Resubscribed to channel"),
             Err(err) => error!(channel, error = ?err, "Failed to resubscribe to channel"),
@@ -219,7 +212,7 @@ impl AsyncClient {
         let shared = Arc::new(Shared {
             interrupt_stream: StdMutex::new(None),
             idle_state: AtomicU8::new(IDLE_STATE_NOT_IDLE),
-            subscribed_channels: StdMutex::new(HashSet::new()),
+            subscribed_channels: SubscribedChannels::default(),
         });
 
         let init =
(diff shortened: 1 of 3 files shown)


Strykar:
@mierak Late note, this merged while I was reading it. The refcounting looks like the
right shape, and I confirmed it flips the case it targets. Config carrying the
same channel twice, otherwise stock:

- 01ee84f (parent): exits 1, "already subscribed to this channel"
- 5b44fc6 (this PR): starts in 183ms

On the bad-plugin case you called out, the repeat part looks fine to me.
Decrement past zero returns None and warns, and if the victim plugin
re-subscribes the 0 to 1 transition re-issues the command, so it recovers.

Separate thing that survives the fix, and I do not think it lands in the same
bucket. MPD caps a single client at 16 channel subscriptions, and rmpcd is one
client for every plugin plus the config list, so that budget is shared. Probed
against MPD 0.24.0:

    subscribe #17 -> ACK [56@0] {subscribe} subscription list is full

Past 16 distinct channels the `mpd.subscribe(channel.clone()).await?` loop in
main.rs aborts startup:

    Error: MpdError: 'Cannot execute command: 'subscribe'. Detail:
    'subscription list is full'. Reason: 'resource already exists'.

(17 entries in `subscribed_channels`, master, otherwise stock config, exit 1.)

Two differences from the bad-plugin case. Nothing has to misbehave, 17 distinct
channels declared exactly as documented is enough. And it takes the whole
daemon down at startup instead of degrading one channel, so every well-behaved
plugin and MPRIS go with it. It is the same abort this PR removed for
duplicates, reached by a different route. Not urgent at four builtins, but the
package manager makes the ceiling reachable, and the raw ACK does not say which
plugin to drop.

Would you rather it keep failing hard with a message naming the channel and the
cap, or warn and skip past 16 so the daemon still comes up?

Two smaller things.

`saturating_sub` in `decrement` cannot saturate. Zero-count entries get
removed, so anything `get_mut` finds is already at least 1. Swapping it for
`*count -= 1`, which panics on underflow in debug, leaves the suite green.
Related, `decrement_below_zero_returns_none` does not exercise that, it
exercises the `None` from the removed key.

`resubscribe()` logs a failed subscribe and keeps the refcount, so a later
`mpd.subscribe()` short-circuits to Ok instead of re-issuing the command and
repairing it, which is what would have happened before this PR. I could not
reach the bad state though. The reachable failure is the socket dying
mid-resubscribe, and that self-heals because the next `enter_idle` fails and
the reconnect cycle repeats. Flagging it as a masked error path rather than a
bug.

mierak:
> MPD caps a single client at 16 channel subscriptions, and rmpcd is one
client for every plugin plus the config list, so that budget is shared.

I wasn't aware of this limitation. I checked MPD's source and it is indeed the case and its hardcoded as well.
One of three things will need to happen in the future if this ever becomes an issue then:
1. Make a PR to MPD to pull the hardcoded limit to mpd.conf
2. Use a separate MPD connection per plugin (probably the best, though feels icky to me)
3. Do not allow plugins to subscribe to channels and instead use a single shared channel and route the messages somehow on app level (probably the worst option)

But I think it is ok to leave as is for now, 16 is still quite a few.

> Would you rather it keep failing hard with a message naming the channel and the
cap

A failure to start is imho better than a silent failure and an ok startup

> saturating_sub in decrement cannot saturate.

I am aware, mostly future proofing against my own dumb mistakes in the future, the perf hit is negligible.

> resubscribe() logs a failed subscribe and keeps the refcount, so a later
mpd.subscribe() short-circuits to Ok

This has to be this way I think. The refcount is stort of a "desired state", it should not be modified and rmpcd should attempt to reach it which should happen under the current conditions.

PR #1091, opened 2026-09-05
fix: album art double rendering on config reload

mierak:
## Description

@NotMephisto sorry for ping, this should be yet another, this time a small improvement to the config reload, might help the stale, lingering, duplicated image being still shown that was on your video too 

### Checklist

- [x] All tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@ keybinds
 - Sixel image disappearing when switching tmux panes
 - Improve responsivness when cava cava is starting/quitting, especially when reloading config
 - MPD's error no longer parsing command index into u8 which has insufficient space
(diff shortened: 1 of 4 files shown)


NotMephisto:
> @NotMephisto sorry for ping, this should be yet another, this time a small improvement to the config reload, might help the stale, lingering, duplicated image being still shown that was on your video too

No worries at all about the ping! It now loads 2 times instead of 4: first when changing the track, and second when changing the theme

mierak:
Thanks! Thats about as good as its gonna get for this use case I think though. I cant think of a way to resolve this save for implementing the color changes directly into rmpc which is not exactly on the radar, not for now anyway.

Still a good improvement though, thanks for testing it!

PR #1096, opened 2026-09-10
fix(nix): remove dependency on flake-utils and use channels tarball

itsyunaya:
## Description
`flake-utils` is used for generating per-platform package and devshell outputs via its `lib.eachDefaultSystem` function. Since that's the only use of the library in the flake, it can be replaced with a small helper function, resulting in virtually the same output but without pulling in another dependency.

Additionally, the inputs URL to Nixpkgs has been replaced with one pointing to `channels.nixos.org`, as this reduces reliance on Github and results in a smaller tarball size. For more info, see [this guide](https://discourse.nixos.org/t/psa-use-nixos-org-tarballs-for-your-flake-inputs/79950) on the NixOS discourse page.

### Related issues
None

### Checklist

- [ ] All tests passed


mierak:
@Kladki can you please take a look?

Kladki:
Honestly, I personally wouldn't bother making either of these changes, since the size difference is insigifnicant compared to the gigabytes your system will be total. But anyways, I guess the system changes could be accepted now, but I think the nixpkgs input url change is kind of jumping the gun if that is still the default for both flakes and install media.

itsyunaya:
> but I think the nixpkgs input url change is kind of jumping the gun if that is still the default for both flakes and install media.

It's an emergent standard that various large projects like `nvf` or `nh` have already incorporated into their Flakes, and a PR to update the template generated by `nix flake init` has gained approval from multiple maintainers as well. I agree that such a change should be handled with caution, but in this case there are no potential downsides to making it.

Kladki:
Well it's now the default for flakes, so I'll give this the all-clear to me merged then, thank you.

mierak:
Alright, thank you both!

